PR #13, opened 2018-04-21
009 integration testing

wkozyra95:
it's not full test suite but i think it's important to merge this now to stop blocking other tasks


wkozyra95:
checks won't pass you can ignore them for now

PR #53, opened 2021-09-02
Demo of django app

grzanka:


grzanka:
@p1003 take a look on this

PR #77, opened 2021-10-09
066 add automatic deploy

### LINKED ISSUE #66: Add automatic deploy of testing version of the backend

retkiewi:
Resolves #66 
Added a deployment job to the CI pipeline


requirements.txt
```diff
@@ -2,4 +2,8 @@ flask
 flask-sqlalchemy
 flask-restful
 pytest
-environs
\ No newline at end of file
+<<<<<<< HEAD
```
grzanka: Fix incorrect merge

tests/test_main.py
```diff
@@ -24,3 +24,12 @@ def test_app_started(client):
     resp = client.get("/")
 
     assert resp.json['message'] == 'Hello world!'
+<<<<<<< HEAD
```
grzanka: Fix incorrect merge

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,60 @@
...
+      - name: Build and run the app via ssh
+        # if: github.ref == 'refs/heads/master'
+      - uses: appleboy/ssh-action@master
+        with:
+          host: ubuntu@30.30.30.8
+          key: ${{ secrets.DEPLOY_C3_TEST }}
+          proxy_host: ubuntu@149.156.182.181
+          script: |
+            cd /yaptide
+            python3 -m pip install --upgrade pip
```
grzanka, line 94: This should work in venv !

grzanka (changes requested):

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,60 @@
...
+            pip3 install -r requirements.txt
+            FLASK_ENV=dev FLASK_APP=yaptide.application flask run
```
grzanka: What if some `flash run` is already running ?

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,60 @@
...
+
+      #send it to the openstack machine via scp
+      - name: Copy files to machine
+        # if: github.ref == 'refs/heads/master'
+        run: |
+          scp -oProxyJump="ubuntu@149.156.182.181" -i ${{ secrets.DEPLOY_C3_TEST }} -r ..\yaptide ubuntu@30.30.30.8:~
+
+      - name: Build and run the app via ssh
+        # if: github.ref == 'refs/heads/master'
+      - uses: appleboy/ssh-action@master
```
grzanka:
```suggestion
      uses: appleboy/ssh-action@master
```

I suspect this should belong to a step with name above

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,60 @@
...
+    steps:
+      #get the repo
+      - uses: actions/checkout@v2
+      - name: Checkout
+
+      #send it to the openstack machine via scp
+      - name: Copy files to machine
+        # if: github.ref == 'refs/heads/master'
+        run: |
+          scp -oProxyJump="ubuntu@149.156.182.181" -i ${{ secrets.DEPLOY_C3_TEST }} -r ..\yaptide ubuntu@30.30.30.8:~
```
grzanka: Will it work ?
I am not sure, but how does github action keep secrets?
I suspect they are like strings or environmental variables, not files.
On the other hand `scp` (like `ssh`) expects to see a filename after `-i` option.

To solve this problem you may use an ssh agent, like here:
https://github.com/yaptide/ui/blob/master/.github/workflows/node.js.yml#L47

retkiewi: I suspected this would not work, but wanted to make sure. I was not sure how to solve this securely.

.github/workflows/test-run.yml
retkiewi: I misunderstood the puprose of the "-", there are a few simmilar errors.

Fixed along with the other errors in b66db34d23825b5daf619e586ef1f6ba1cc16003, also introduced a more consistent step structure.

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,62 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+      
+      - name: Setup Python
         uses: actions/setup-python@v2
         with:
           python-version: "3.8"
+          
+      #install requirements.txt
       - name: Instal dependencies
```
grzanka:
```suggestion
      - name: Install dependencies
```

.github/workflows/test-run.yml
retkiewi: Fixed in 7cc44d7adb33d077f0f7719a59d331d0f29abe1d

.github/workflows/test-run.yml
retkiewi: The command will return an error stating that the port used by flask is already in use.

grzanka: Then we should kill the flask process and run it from scratch

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,72 @@
...
+      - name: Build and run the app via ssh
+        # if: github.ref == 'refs/heads/master'
+        uses: appleboy/ssh-action@master
+        with:
+          host: ubuntu@30.30.30.8
+          key: ${{ secrets.DEPLOY_C3_TEST }}
+          proxy_host: ubuntu@149.156.182.181
+          script: |
+            cd /yaptide
+            python3 -m venv env
```
grzanka, line 92: What if venv in `env` directory exists ?

retkiewi: Since we don't run venv with the --clear option, the venv stays intact if present, but if absent a new venv is created.

grzanka: ok

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,76 @@
...
+#           proxy_host: "149.156.182.181"
+#           proxy_username: ubuntu
+#           proxy_key: ${{ secrets.DEPLOY_C3_TEST }}
+#           host: "30.30.30.8"
+#           username: ubuntu
+#           key: ${{ secrets.DEPLOY_C3_TEST }}
+#           source: "./*"
+#           target: "/scptest"
+
+      - name: Copy files to machine
```
grzanka: What about this one which uses rsync instead of scp ?

rsync should accept additional options to copy via intermedia host 
https://github.com/marketplace/actions/ssh-deploy
https://superuser.com/questions/1115715/rsync-files-via-intermediate-host

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,76 @@
...
+
+  deploy:
+    runs-on: ubuntu-latest
+    needs: [build-and-test]
+    steps:
+      #get the repo
+      - name: Checkout
+        uses: actions/checkout@v2
+    
+#       - name: Copy files to machine
```
grzanka: For debugging consider using this one https://github.com/marketplace/actions/wireguard-ssh

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,76 @@
...
+# #         if: github.ref == 'refs/heads/master'
```
grzanka: If I understand correctly copying the key to the file can be done with this one:
https://github.com/marketplace/actions/install-ssh-key

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,76 @@
...
+#           proxy_key: ${{ secrets.DEPLOY_C3_TEST }}
+#           host: "30.30.30.8"
+#           username: ubuntu
+#           key: ${{ secrets.DEPLOY_C3_TEST }}
+#           source: "./*"
+#           target: "/scptest"
+
+      - name: Copy files to machine
+        run: |
+          echo "${{ secrets.DEPLOY_C3_TEST }}" | scp -i /dev/stdin -oProxyJump="ubuntu@149.156.182.181" -r ./* ubuntu@30.30.30.8:~/scptest
```
grzanka: Interesting hint from https://github.com/marketplace/actions/install-ssh-key
```SCP is deprecated by OpenSSH due to outdated and inflexible protocol.```
Read more here https://www.openssh.com/txt/release-8.0
```
The scp protocol is outdated, inflexible and not readily fixed. We
recommend the use of more modern protocols like sftp and rsync for
file transfer instead.```

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,71 @@
...
+    needs: [build-and-test]
+    steps:
+      #get the repo
+      - name: Checkout
+        uses: actions/checkout@v2
+    
+      - name: set known_hosts env variable
+        run: expo
+      
+      - name: Install ssh key
```
grzanka: It is not needed to add ssh key to the file, works well with the agent, as shown here:
https://github.com/yaptide/yaptide/pull/82/files

then only disadvantage of my solution is to disable host checking at proxy server

.github/workflows/test-run.yml
```diff
@@ -1,36 +1,84 @@
...
+          proxy_host: 149.156.182.181
+          proxy_username: ubuntu
+          proxy_key: ${{ secrets.DEPLOY_C3_TEST }}
+          script: |
+            cd yaptide
+            python3 -m venv env
+            source env/bin/activate
+            python3 -m pip install --upgrade pip
+            pip3 install -r requirements.txt
+            FLASK_ENV=developement FLASK_APP=yaptide.application nohup flask run >> /dev/null &
```
grzanka: Use `screen` instead o f `nohup`.
Straightforward solution described here https://unix.stackexchange.com/questions/266571/remotely-starting-a-screen-session-through-ssh-and-closing-the-ssh-session-immed

Tested on my system:
image

To start htop remotely, keep it running and disconnect I type:
```
ssh ubuntu@149.156.182.181 screen -d -m "htop"
```

To check if it is still running, after a while I connect to the machine and then connect to a running `screen` session:
```
ssh ubuntu@149.156.182.181
screen -x
```

Get familiar with `screen` - it is a great tool

retkiewi: I've swapped out the ```nohup``` + ```& exit``` for ```screen```.

.github/workflows/test-run.yml
retkiewi: I've swapped the install-ssh action for the ssh-agent action and added the known_hosts file via a script inspired by your solution. 

Regarding the disadvantage, can't we just append the known_hosts file with the output from ssh-keyscan ran from the external machine?
(something like ```ssh <external machine address> 'ssh-keyscan -H <internal machine address>' >> .ssh/known_hosts```)

.github/workflows/test-run.yml
retkiewi: I've removed SCP in favour of rsync.

grzanka (approved):
Looks OK

PR #103, opened 2021-10-30
Add converter to Flask App

p1003:


yaptide/routes.py
```diff
@@ -44,13 +44,18 @@ def get():
         errors = shschema.validate(args)
         if errors:
             return errors
-
         param_dict = shschema.load(args)
-        simulation_result = run_shieldhit(param_dict)
+
+        json_data = request.json
+        if not json_data:
+            return {"status": "json error"}
```
retkiewi: Add appropriate rest response codes.
https://documentation.commvault.com/11.24/essential/45599_rest_api_response_codes_and_statuses.html

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,95 +10,31 @@
...
-ICRU 276
-END
-"""
-input_cfg_templ['detect.dat'] = """
-Geometry Cyl
-    Name ScoringCylinder
-    R 0.0 10.0 {cyl_nr:d}
-    Z 0.0 30.0 {cyl_nz:d}
+from ..converter.converter.converter import DummmyParser
+from ..converter.converter.converter import Runner as ConvertRunner
```
retkiewi: `ConvertRunner` seems wierd, wouldn't either `ConvRunner` or `ConverterRunner` be better? Or you could even call it `Converter`.

yaptide/routes.py
```diff
@@ -44,13 +44,18 @@ def get():
...
+
+        simulation_result = run_shieldhit(param_dict=param_dict,
+                                          json_to_convert=json_data)
 
         if simulation_result:
             return {"status": "ok"}
-        return {"status": "error"}
+        return {"status": "sim error"}
```
retkiewi: I would either go with `sim_err` or `simulation error`.

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,95 +10,31 @@
...
-  003          +3     -2
-  END
-* material codes: 1 - liquid water (ICRU material no 276), 1000 - vacuum, 0 - black body
-    1    2    3
-    1 1000    0
-"""
-
-
-def run_shieldhit(param_dict):
+def run_shieldhit(param_dict, json_to_convert):
```
retkiewi: Change `json_to_convert` to something else (eg. input_dict). `json_to_convert` suggests that it is the runners responsibility to convert the input. It is also confusing that the runner has no explicit input data.

Also consider changing `param_dict` to `**kwargs`.

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,95 +10,31 @@
...
 
-        runner_obj = Runner(jobs=param_dict['jobs'],
-                            keep_workspace_after_run=False,
-                            output_directory=tmp_output_path)
+        runner_obj = SHRunner(jobs=param_dict['jobs'],
+                              keep_workspace_after_run=False,
+                              output_directory=tmp_output_path)
 
         start_time = timeit.default_timer()
         isRunOk = runner_obj.run(settings=settings)
```
retkiewi, line 42: Why create the `isRunOk` variable instead of just doing: `if not runner_obj.run(settings=settings): return None`

yaptide/routes.py
p1003:
```suggestion
            return {"status": 400}
```

yaptide/simulation_runner/shieldhit_runner.py
p1003: Changed as suggested

yaptide/simulation_runner/shieldhit_runner.py
p1003: Used `json_to_conver`

yaptide/routes.py
p1003: Changed

yaptide/simulation_runner/shieldhit_runner.py
p1003: Changed

yaptide/simulation_runner/shieldhit_runner.py
retkiewi: I don't see how introducing a typo would make this more readable.

yaptide/routes.py
retkiewi: That's not what i meant. Look at other endpoints and [this guide](https://github.com/flask-api/flask-api/blob/develop/docs/api-guide/status-codes.md).

yaptide/routes.py
```diff
@@ -37,20 +38,28 @@ class ShieldhitDemo(Resource):
...
         errors = shschema.validate(args)
         if errors:
             return errors
-
         param_dict = shschema.load(args)
-        simulation_result = run_shieldhit(param_dict)
+
+        json_data = request.json
+        if not json_data:
+            return ({"message": "Input Json Error"},
```
retkiewi, line 52: Why are you returning tuples? Also, you can just return the codes as numbers.

yaptide/routes.py
p1003: @retkiewi check it know please

yaptide/routes.py
```diff
@@ -37,20 +38,28 @@ class ShieldhitDemo(Resource):
...
+                    api_status.HTTP_400_BAD_REQUEST)
```
retkiewi, line 53: Why is this in a new line? The line has less than 120 characters.

yaptide/routes.py
p1003: If I send some request I would like to know what went wrong. Tuple is used to provide info about it. Talking about returning status codes it is just a way of doing it.

yaptide/routes.py
```diff
@@ -37,20 +38,28 @@ class ShieldhitDemo(Resource):
     """Class responsible for Shieldhit Demo running"""
 
     @staticmethod
-    def get():
+    def post():
```
grzanka, line 41: I would like to test it with curl,
add some description how to test it with curl (command line)

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,39 @@
...
-    )
-
     # create temporary directory
     with tempfile.TemporaryDirectory() as tmp_output_path:
 
-        for config_filename in input_dict:
-            abs_input_path = os.path.join(tmp_output_path, config_filename)
-            with open(abs_input_path, 'w') as temp_input_file:
-                temp_input_file.write(input_dict[config_filename])
+        conv_runner = ConvRunner(parser=DummmyParser(),
```
grzanka, line 26:
```suggestion
        # digest dictionary with project data (extracted from JSON file) and generate SHIELD-HIT12A input files
        conv_runner = ConvRunner(parser=DummmyParser(),
```

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,39 @@
...
+                                 input_data=raw_input_dict,
+                                 output_dir=tmp_output_path)
+
+        conv_runner.run_parser()
 
+        print(os.listdir(tmp_output_path))
```
grzanka:
```suggestion
```

Not needed here.

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,39 @@
...
-                            keep_workspace_after_run=False,
-                            output_directory=tmp_output_path)
+        runner_obj = SHRunner(jobs=param_dict['jobs'],
+                              keep_workspace_after_run=False,
+                              output_directory=tmp_output_path)
 
         start_time = timeit.default_timer()
-        isRunOk = runner_obj.run(settings=settings)
-        if not isRunOk:
+        if not runner_obj.run(settings=settings):
```
grzanka: This is a bad idea. Working with debugger is much easier if status code is extracted into separate value, then this value is used in `if` statement

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,39 @@
...
             return None
 
         elapsed = timeit.default_timer() - start_time
         print("MC simulation took {:.3f} seconds".format(elapsed))
 
         estimator = runner_obj.get_data()
```
grzanka: Is this really single `estimator` object as the name of the variable suggest or a dictionary with keys being a corenames of the output files and values being objects of estimator class ?

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,39 @@
...
 
-        return estimator
-
-
-if __name__ == '__main__':
-    sys.exit(run_shieldhit(sys.argv[1:]))
+        return estimator['mesh_']
```
grzanka: We have now configuration generated from JSON file, corenames of the output files may largely differ, depending on what user provides in the JSON file. Why do you assume a fixed value of `mesh_` ?

grzanka (changes requested):
see comments

yaptide/simulation_runner/shieldhit_runner.py
grzanka: Wrong suggestion. Imagine working with debugger. How would you add a breakpoint on the exit code ?

yaptide/simulation_runner/shieldhit_runner.py
p1003: Yes it should be deleted, but I had forgotten about it

yaptide/simulation_runner/shieldhit_runner.py
p1003: It is changed in #104

yaptide/simulation_runner/shieldhit_runner.py
p1003: changed in #104

yaptide/routes.py
p1003: Currently dummy converter ignores the content of json in requests body so it can contains anything. Example curl for Windows:
`curl -i -X POST -H "Content-Type:application/json" -d "{\"Dummy\": \"Curl\" }" http://localhost:5000/sh/demo `
And for Linux:
`curl -i -X POST -H "Content-Type:application/json" -d '{"Dummy": "Curl" }' "http://localhost:5000/sh/demo"`

PR #104, opened 2021-10-30
Return simulation results

### LINKED ISSUE #90: Digest and send back results of SH12A calculation
Right now all calls to `/sh/demo` gives back status : ok.
Instead I would expect to get all ingredients needed to present results to the user (i.e. in a plot).

p1003:


yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -92,26 +93,68 @@ def run_shieldhit(param_dict):
...
         print("MC simulation took {:.3f} seconds".format(elapsed))
 
-        estimator = runner_obj.get_data()
-
-        return estimator
-
-
-if __name__ == '__main__':
-    sys.exit(run_shieldhit(sys.argv[1:]))
+        estimator_list = runner_obj.get_data()
```
grzanka: This is misleading, `get_data` is returning a dictionary, not a list, see:
https://github.com/DataMedSci/pymchelper/blob/master/pymchelper/executor/runner.py#L89

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -92,26 +93,68 @@ def run_shieldhit(param_dict):
...
+
+        return convert_output(estimator_list)
+
+
+def convert_output(estimator_list):
+    """Function for converint simulation output to dictionary"""
+    if not estimator_list:
+        return {"result": "None"}
+
+    result_dict = {"estimators": []}
```
grzanka, line 58: What is the point in creating a dictionary with single key "estimator" ?
In principle this method could simply return a list.

General remark - add comment lines explaining your architectural choices.
I suspect that this strange dictionary may be dictated by later need to use in some JSON stuff.
If that is the case - write it down, explaining the reasons, not the details of the implementation.

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -92,26 +93,68 @@ def run_shieldhit(param_dict):
...
+    for est_name in estimator_list:
```
grzanka: This line suggest that `estimator_list` is a collection of strings being estimator names.
Better name would be `est_object` or simply `estimator`

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -92,26 +93,68 @@ def run_shieldhit(param_dict):
...
+
+        estimator_dict = {
+            "name" : est_name,
+            "pages": []}
+        for page in estimator_list[est_name].pages:
```
grzanka: What if the data dimension is different than one ?
Code below has several flaws:
 - it violates DRY principle
 - it doesn't work for 0-D data (a single number) - in such case it is adding same number 5 times
 - it doesn't work for 2-D data

If you are planning to add a simple version of the converter which works with 1-D data, then I suggest to:
 - use `page.dimension` property (I assume you know how to access python property) to check if data dimension is 1, see https://github.com/DataMedSci/pymchelper/blob/master/pymchelper/page.py#L48
 - in case dimension is different than 1, print some warning and return nothing
 - in case dimension is 1, the use `page.plot_axis(0)` to get relevant axis. Before doing so, read carefully docstring and code of this method https://github.com/DataMedSci/pymchelper/blob/master/pymchelper/page.py#L123
 - for data extraction you can use `page.data_raw` or `page.data.flatten()` (assuming you know what numpy flatten is doing). For inspiration check how converter to PNG files is working: https://github.com/DataMedSci/pymchelper/blob/master/pymchelper/writers/plots.py#L208

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -92,26 +93,68 @@ def run_shieldhit(param_dict):
...
+            page_dict = {"axis": []}
+            for i in range(5):
+                axis_dict = {
+                    "n": int(page.axis(i).n),
+                    "min_val": float(page.axis(i).min_val),
+                    "max_val": float(page.axis(i).max_val),
+                    "name": str(page.axis(i).name),
+                    "unit": str(page.axis(i).unit),
+                    "binning": str(page.axis(i).binning),
+                    "data": []
```
grzanka: Why don't you extract X axis data as well ?
See `axis.data` property (not to be confused with `page.data` property
https://github.com/DataMedSci/pymchelper/blob/master/pymchelper/axis.py#L40

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -92,26 +93,68 @@ def run_shieldhit(param_dict):
...
+                elif i == 2:
+                    for val in page.data[0, 0, :, 0, 0]:
+                        axis_dict["data"].append(float(val))
+                elif i == 3:
+                    for val in page.data[0, 0, 0, :, 0]:
+                        axis_dict["data"].append(float(val))
+                elif i == 4:
+                    for val in page.data[0, 0, 0, 0, :]:
+                        axis_dict["data"].append(float(val))
+                page_dict["axis"].append(axis_dict)
```
grzanka: Do you really want to pass all 5 pages to frontend ?
That means that frontend would need to have the logic needed to decide whether data to present is a number, 1D plot or 2D plot. This is a bad practice (too much logic on frontend).

grzanka (changes requested):
see comments

yaptide/simulation_runner/shieldhit_runner.py
p1003: My mistake

yaptide/simulation_runner/shieldhit_runner.py
p1003: Updated

yaptide/application.py
```diff
@@ -17,6 +17,15 @@ def create_app(config_object="yaptide.settings"):
 
     api = Api(app)
     initialize_routes(api)
+
+    @app.after_request
+    def after_request(response):
+        response.headers.add('Access-Control-Allow-Origin', '*')
```
grzanka: Why this is needed?

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,91 @@
...
+    """Dummy function for converting simulation output to dictionary"""
+    if not estimators_dict:
+        return {"message": "No estimators"}
+
+    # result_dict is the dictionary object, which is later converted to json
+    # to provide readable api response for fronted
+
+    # result_dict contains the list of estimators
+    result_dict = {"estimators": []}
+    for estimator in estimators_dict:
```
grzanka: From this code I read that you read from a dictionary and you assume that the keys are objects of the `Estimator` class (I deduce it from object name `estimator`).
I suspect that is not the case, `estimators_dict` has keys being strings, hence `estimator` variable name is very misleading

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,91 @@
...
+
+        # estimator_dict contains list of pages
+        estimator_dict = {
```
grzanka: You have here `estimators_dict` and `estimator_dict`. Both have very similar name, but completely different structure.
Use better naming.

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,91 @@
...
+                y_values = page.data_raw.flatten()
+
+                # for 1 dimension page, dict contains:
+                # "dimensions" indicating it is 1 dim page
+                # "unit"
+                # "x_y_pairs" which is list of x any y values pairs
+                page_dict = {
+                    "dimensions" : page_dim,
+                    "unit": str(axis.unit),
+                    "x_y_pairs": []
```
grzanka: Why are you creating list of (x,y) pairs ?
Looking at:
image
https://github.com/yaptide/ui/pull/226/files#diff-58cddae6837f8e66ec955968adbef3da86fb51763ec55e990dce8398b3d0131cR23
it seems that frontend prefers list of x and list of y values, separately.

grzanka (changes requested):
see comments

yaptide/simulation_runner/shieldhit_runner.py
p1003: I have no idea for better naming, can you suggest something?

grzanka:
Also - it doesn't work on backend
image

yaptide/application.py
p1003: This solves the problem with CORS, which occured during connecting back with front

grzanka: add a comment with detailed explanation

yaptide/simulation_runner/shieldhit_runner.py
p1003: Changed

yaptide/simulation_runner/shieldhit_runner.py
grzanka: I'd suggest to get both key and value in same way:
```
for estimator_name, estimator_obj in estimators_dict.items():
```

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,91 @@
...
+
+                # for 1 dimension page, dict contains:
+                # "dimensions" indicating it is 1 dim page
+                # "unit"
+                # "x_values" which is list of x values
+                # "y_values" which is list of y values
+                page_dict = {
+                    "dimensions" : page_dim,
+                    "unit": str(axis.unit),
+                    "x_values": [],
```
grzanka: Why not simply:
```suggestion
                    "x_values": axis.data,
```
?

Same for y_values

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,91 @@
...
+    """Dummy function for converting simulation output to dictionary"""
+    if not estimators_dict:
+        return {"message": "No estimators"}
+
+    # result_dict is the dictionary object, which is later converted to json
+    # to provide readable api response for fronted
+
+    # result_dict contains the list of estimators
+    result_dict = {"estimators": []}
+    for est_name in estimators_dict:
```
grzanka: Take a look here again:
https://github.com/yaptide/yaptide/pull/104#discussion_r740534633

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,91 @@
...
+        # est_dict contains list of pages
+        est_dict = {
+            "name" : est_name,
+            "pages": []}
+        for page in estimators_dict[est_name].pages:
+
+            page_dim = page.dimension
+
+            # handling 1 dimension page
+            if page_dim == 1:
```
grzanka:
```suggestion
            # handling 1 dimension page
            if page.dimension == 1:
```

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,91 @@
...
+                y_values = page.data_raw.flatten()
+
+                # for 1 dimension page, dict contains:
+                # "dimensions" indicating it is 1 dim page
+                # "unit"
+                # "x_values" which is list of x values
+                # "y_values" which is list of y values
+                page_dict = {
+                    "dimensions" : page_dim,
+                    "unit": str(axis.unit),
```
grzanka: What about axis name ?
How frontend should know whether we want to have X,Y or Z on horizontal axis ?

yaptide/simulation_runner/shieldhit_runner.py
p1003: > I'd suggest to get both key and value in same way:
> 
> ```
> for estimator_name, estimator_obj in estimators_dict.items():
> ```

Changed like suggested in this comment

yaptide/simulation_runner/shieldhit_runner.py
p1003: The problem I found was that, the axis didn't actually have any name (string was empty) but we will discuss it on the meeting

grzanka: Even if the name is empty, use it. Fix in pymchelper is incoming https://github.com/DataMedSci/pymchelper/pull/477

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,84 @@
...
+                # for 1 dimension page, dict contains:
+                # "dimensions" indicating it is 1 dim page
+                # "unit"
+                # "x_values" which is list of x values
+                # "y_values" which is list of y values
+                page_dict = {
+                    "dimensions" : page.dimension,
+                    "unit": str(axis.unit),
+                    "x_values": axis.data,
+                    "y_values": page.data_raw.flatten()
```
grzanka: Result units and names are missing (y unit and y name), probably in `page.name` and `page.unit`

grzanka: Consider data structure as value + name + unit

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,84 @@
...
+    for estimator_name, estimator_obj in estimators_dict.items():
+
+        # est_dict contains list of pages
+        est_dict = {
+            "name" : estimator_name,
+            "pages": []}
+        for page in estimator_obj.pages:
+
+            # handling 1 dimension page
+            if page.dimension == 1:
```
grzanka, line 78: In case of higher number of dimensions, return number of dimensions.
Later we will extend that data with path to the output file (when simulation will be running on SLURM)

grzanka:
@p1003 I suggest rebase onto newest master

grzanka:
@p1003 due to recent changes in deploy, I'd recommend another rebase

yaptide/simulation_runner/shieldhit_runner.py
grzanka: Fixed in 1.9.1 https://github.com/DataMedSci/pymchelper/releases

yaptide/simulation_runner/shieldhit_runner.py
p1003: Changed

yaptide/simulation_runner/shieldhit_runner.py
p1003: Added `name`

yaptide/application.py
```diff
@@ -17,6 +17,15 @@ def create_app(config_object="yaptide.settings"):
...
     api = Api(app)
     initialize_routes(api)
+
+    # @app.after_request
+    # def after_request(response):
+    #     response.headers.add('Access-Control-Allow-Origin', '*')
+    #     response.headers.add('Access-Control-Allow-Headers',
+    #                          'Content-Type,Authorization')
+    #     response.headers.add('Access-Control-Allow-Methods', 'GET,PUT,POST,DELETE')
+    #     return response
```
grzanka:
```suggestion
```

remove

yaptide/routes.py
```diff
@@ -1,4 +1,5 @@
-from flask import request
+from flask import request, json
+from flask_api import status as a_s
```
grzanka: why `a_s` ? use some more verbose name

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,98 @@
...
+                              output_directory=tmp_output_path)
 
         start_time = timeit.default_timer()
+
         isRunOk = runner_obj.run(settings=settings)
         if not isRunOk:
             return None
 
         elapsed = timeit.default_timer() - start_time
         print("MC simulation took {:.3f} seconds".format(elapsed))
```
grzanka: remove this printout

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,98 @@
...
+                    "values": axis.data.tolist()
+                }
+            # if page.dimension > 0:
+            # if page.dimension > 1:
+            #     axis = page.plot_axis(1)
+            #     page_dict["second_axis"] = {
+            #         "unit": str(axis.unit),
+            #         "name": str(axis.name),
+            #         "values": axis.data.tolist()
+            #     }
```
grzanka: remove
```suggestion
```

yaptide/simulation_runner/shieldhit_runner.py
```diff
@@ -10,108 +10,98 @@
...
+                page_dict["data"] = {
+                    "unit": str(page.unit),
+                    "name": str(page.name),
+                    "values": page.data_raw.flatten().tolist()
+                }
+                axis = page.plot_axis(0)
+                page_dict["first_axis"] = {
+                    "unit": str(axis.unit),
+                    "name": str(axis.name),
+                    "values": axis.data.tolist()
```
grzanka, line 88: why `tolist` ?

p1003: Np arrays are not json serializable

README.md
```diff
@@ -33,3 +33,17 @@ When you're ready to stop the container, use the following command:
 docker stop yaptide
 docker rm yaptide
 ```
+
+# Curl
```
grzanka:
```suggestion
# Testing API with command-line tools
```

README.md
```diff
@@ -33,3 +33,17 @@ When you're ready to stop the container, use the following command:
...
+
+Currently dummy converter ignores the content of json sent in request's body so it can contains anything.
```
grzanka:
```suggestion
Currently, the dummy converter ignores the JSON content sent in the request's body so it can contain anything.
```

yaptide/simulation_runner/shieldhit_runner.py
grzanka: great, thanks

yaptide/routes.py
p1003: changed

yaptide/simulation_runner/shieldhit_runner.py
p1003: removed

grzanka (dismissed):
looks good

grzanka (approved):
good

PR #112, opened 2021-11-02
Grzanka/replace-python-deployment-106

### LINKED ISSUE #106: Replace python deployment with Docker
Backend VM currently used for testing has installed python 3.8.
We need python version at least 3.9, in order to have working converter.
Dockerfile has python 3.9, so we could use it.
The only issue to solve is installation of shieldhit binary inside docker.

This can be done by extending Dockerfile with `ARG` entries, like here:
https://github.com/flukadocker/F4D/blob/master/common/fluka.dockerfile

The purpose of this `ARG` entries is to provide location of executable file(s) to be copied inside docker.
THey can go to `/usr/local/bin` or some other dedicated directory in PATH

grzanka:


.github/workflows/test-run.yml
```diff
@@ -99,11 +102,10 @@ jobs:
...
+          command_timeout: 30m
           script: |
             cd yaptide
-            rm -rf venv
-            python3 -m venv env
-            source env/bin/activate
-            python3 -m pip install --upgrade pip
-            pip3 install -r requirements.txt
-            FLASK_ENV=developement FLASK_APP=yaptide.application PATH=$PATH:${{ env.SHIELDHIT_PATH }} screen -d -m flask run --host 0.0.0.0
+            cp -r /home/ubuntu/usr/shieldhit ./shieldhit/
```
grzanka: Can you copy single executable instead of directory ?
The same to copying inside the docker

grzanka:
```suggestion
            cp -r /home/ubuntu/usr/shieldhit/v0.9.2/bin/shieldhit .
```

.github/workflows/test-run.yml
```diff
@@ -99,11 +102,10 @@ jobs:
...
+            docker build --build-arg SHIELDHIT_PATH=/shieldhit -t yaptide .
```
grzanka:
```suggestion
            docker build --build-arg SHIELDHIT_PATH=shieldhit -t yaptide .
```

Dockerfile
```diff
@@ -1,9 +1,11 @@
 FROM python:3.9-alpine
+ARG SHIELDHIT_PATH
+COPY $SHIELDHIT_PATH /usr/local/bin/shieldhit
```
grzanka:
```suggestion
COPY $SHIELDHIT_PATH /usr/local/bin/
```

.github/workflows/test-run.yml
```diff
@@ -99,11 +102,10 @@ jobs:
...
+            rm -r ./shieldhit
```
grzanka:
```suggestion
            rm -r shieldhit
```

.github/workflows/test-run.yml
retkiewi: Yep, that's a great idea.

grzanka (commented):
Uncomment master switch

retkiewi (approved):

PR #152, opened 2021-12-07
Cancel simulation and get user simulations

p1003:


yaptide/celery/tasks.py
```diff
@@ -192,3 +192,11 @@ def sh12a_simulation_status(path_to_file: str):
             }
         }
         return sim_info
+
+
+@celery_app.task
+def cancel_simulation(task_id: str) -> dict:
+    """Task responsible for canceling simulation in progress"""
```
grzanka, line 199: Is it sending SIGTEM singal to python process and then to the shieldhit executable ?

grzanka:
Is it possible to obtain output from cancelled simulations ?

yaptide/celery/tasks.py
p1003: No, it does not I need to change it

p1003:
> 

Currenlty it is not and might not be added in this pull request

grzanka (dismissed):
LGTM

yaptide/celery/tasks.py
grzanka: What about changing that in separete PR ?

grzanka (approved):
LGTM

PR #158, opened 2021-12-10
Routes and API response general rework

p1003:


grzanka (dismissed):
LGTM

ostatni5:
Do not send 'No content' as a response. Also, field content is supposed to be a dictionary - keep the same format of fields. 
obraz

obraz

ostatni5:
in /sh/status 
Send in response status of the task. Now is ambiguous if the request failed or the task failed. Because 500 and weird response suggest that the request failed.
obraz

ostatni5 (changes requested):
Adress comments

ostatni5:
What is the purpose of the third cookie it is not removed after logout?
obraz

p1003:
> What is the purpose of the third cookie it is not removed after logout? obraz

It is a bug, `token` shouldn't be present, fixed in next commit

p1003:
> in /sh/status Send in response status of the task. Now is ambiguous if the request failed or the task failed. Because 500 and weird response suggest that the request failed. obraz

This is very hard to manage because it is converter Error but you can assume that this isn't error with your request, because I send you back other status codes if informations in it was incorrect/invalid. You get 500 when server itself or simulation is failing.

yaptide/celery/tasks.py
```diff
@@ -115,57 +115,28 @@ def simulation_task_status(task_id: str) -> dict:
 
     result = {
         'status': 'OK',
-        'message': {
-            'state': task.state
+        'content': {
+            'state': task.state,
```
grzanka, line 121:
```suggestion
            'state': task.state
```

do we need this comma ?

p1003: @grzanka we do not, but this is convention used everywhere in the project as in converter repository.

p1003: I am not commiting this suggestion

grzanka: ok, no problem

grzanka:
@p1003 I leave the review for @ostatni5 , for me it is good enough

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
grzanka: Do we really need to commit that file ?

yaptide/routes/utils/decorators.py
```diff
@@ -0,0 +1,34 @@
...
+    def decorator(f):
+        """Determines if the access or refresh token is valid"""
+        @wraps(f)
+        def wrapper(*args, **kwargs):
+            token: str = request.cookies.get(
+                'refresh_token' if is_refresh else 'access_token')
+            if not token:
+                raise Unauthorized(description="No token provided")
+            resp: Union[int, str] = decode_auth_token(
+                token=token, is_refresh=is_refresh)
```
grzanka:
```suggestion
            resp: Union[int, str] = decode_auth_token(token=token, is_refresh=is_refresh)
```

.vscode/settings.json
p1003: it should have been in gitignore

ostatni5 (dismissed):
I do not see any issues for now

yaptide/routes/utils/decorators.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def requires_auth(is_refresh: bool):
+    """Decorator for auth requirements"""
+    def decorator(f):
+        """Determines if the access or refresh token is valid"""
+        @wraps(f)
+        def wrapper(*args, **kwargs):
+            token: str = request.cookies.get(
+                'refresh_token' if is_refresh else 'access_token')
```
grzanka:
```suggestion
            token: str = request.cookies.get('refresh_token' if is_refresh else 'access_token')
```

grzanka (approved):
LGTM

PR #206, opened 2022-10-06
Fix/simulator binary validation

Derstilon:
Now you can import a shieldhit binary file of any name and it will work just fine.
Local binary files can by stored in ignored directory `./shieldhit/` in repository root
image

Fixed issue with unnecessary files being copied to the docker memory.


grzanka (approved):
LGTM

Issue #233, opened 2022-11-15
Support for new shieldhit binary without gfrotran deps


grzanka:
Up to now SHIELD-HIT12A binary was compiled with `make -j` command. Binary created in such way had a dependency on `gfortran` libraries. Now when I switched to compilation with `make gfortran -j` the binary is fully statically compiled and `gfortran` libraries are not needed anymore.

PR #234, opened 2022-11-15
removing gfrortan deps and python version bump

grzanka:
This PR removed installation of `gfortran` libraries needed by shieldhit binary compiled with `make -j`. Now we switch to a version which is fully statically compiled, see https://github.com/yaptide/yaptide/issues/233#issue-1449420398


grzanka:
I tested this PR with `ansible-playbook site.yml -e backend_repo_version='grzanka/support-for-new-shieldhit-233'` and new shieldhit binary and it seems to work OK.

ostatni5 (approved):
ok

PR #238, opened 2022-11-29
Dummy batch API

p1003:


yaptide/routes/batch_routes.py
```diff
@@ -0,0 +1,152 @@
...
+                cmd_line_args: <słownik> | <None> (ma wygenerować stringa)
+
+        result not ok:
+            "Error message" (scenariusze):
+                error message od połączenia
+                error message batcha
+
+        result ok:
+            job_id
+        """
```
grzanka:
```suggestion
```

move that info to issue description

grzanka:
To me it looks OK, but @Derstilon should check first if everything is there in place

yaptide/routes/batch_routes.py
p1003: Ok

p1003: https://github.com/yaptide/yaptide/issues/242#issuecomment-1430542060
Wrote it down here, consider resolving conversation @grzanka

yaptide/routes/celery_routes.py
```diff
@@ -16,7 +16,7 @@
                                   get_input_files, cancel_simulation)
 
 
-class SimulationRun(Resource):
+class JobsDirect(Resource):
     """Class responsible for SHIELD-HIT12A simulations running"""
```
Derstilon: update description if class is responsible only for simulations in celery directly on backend

Derstilon (changes requested):

Derstilon (dismissed):

Derstilon (dismissed):

Derstilon (approved):

PR #241, opened 2022-12-23
Rewrite rest post requests to use only body

### LINKED ISSUE #240: Rewrite rest post requests to use only body
Linked with [https://github.com/yaptide/ui/issues/759](ui:759)

p1003:


p1003:
List of changes:
1. `/sh/run` endpoint now needs to be fed with following json in body:
```
{
    "jobs": <number>, # optional
    "sim_type": <string>, # optional
    "sim_name": <string>, # optional
    "sim_data": <json> # former json body
}
```
2. Endpoints `/sh/status` and `/sh/inputs` now responds on `get` method not `post` so former JSON bodies should be converted to URL parameters.

yaptide/routes/simulation_routes.py
```diff
@@ -151,18 +146,20 @@ class _Schema(Schema):
 
     @staticmethod
     @requires_auth(is_refresh=False)
-    def post(user: UserModel):
+    def get(user: UserModel):
```
Derstilon, line 149: Do I have to rewrite front end to send get instead of post?

p1003: Yes, you have to change it.

Derstilon (approved):

PR #252, opened 2023-02-06
Database dependecies update and fix

p1003:
image
For now it is not really clear how to deal with the issue.

`autoload` was deprecated and removing this parameter worked fine, but further changes in API caused problem with persistance itself. It seems that adding user has no effect, while DB doesn't throw any error about failure.


grzanka (approved):
LGTM

PR #258, opened 2023-02-17
Celery multiple task progress

p1003:


grzanka:
I think this needs to be tested together with new branch on UI

p1003:
> I think this needs to be tested together with new branch on UI

There are important changes in the API so it has to be tested together

README.md
```diff
@@ -126,11 +129,11 @@ And for Linux:
 curl -i -X POST -b cookies.txt -H "Content-Type:application/json" -d @path/to/jsonfile "http://localhost:5000/sh/run"
 ```
 
-You can also add parameters after ``?`` sign like this: ``http://localhost:5000/sh/run?<param name>=<param value>``
+You can also add parameters after `?` sign like this: `http://localhost:5000/sh/run?<param name>=<param value>`
```
grzanka: Do we still use `sh/run` ?

p1003: no

grzanka (approved):
LGTM

PR #272, opened 2023-03-09
Fix FileNotFound Error

### LINKED ISSUE #267: FileNotFoundError in Celery
It seems that there are some situations when temporary working directory is deleted. But result is not present in the response. It might be caused by asynchronous architecture so should be further analysed and handled more properly. Suspect: at the moment of requesting job status it is at the very begining but still RUNNING and when the operations required for proper checking of RUNNING job are started, the job is already COMPLETED, but this information in current job status check is unavailable.
```
File "/usr/local/app/yaptide/celery/tasks.py", line 246, in sh12a_simulation_status
    for workdir in os.listdir(dir_path):
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpegygkgt8'
```

p1003:


yaptide/celery/tasks.py
```diff
@@ -243,17 +243,18 @@ def sh12a_simulation_status(dir_path: str, sim_ended: bool = False) -> list:
     """Extracts current SHIELD-HIT12A simulation state from first available logfile"""
     # This is dummy version because pymchelper currently doesn't privide any information about progress
     result_list = []
-    for workdir in os.listdir(dir_path):
-        if not re.search(r"run_", workdir):
+    dir_path: Path = Path(dir_path)
+    if not dir_path.exists():
+        return result_list
+    for workdir in dir_path.iterdir():
```
grzanka: Consider glob expression in form of `workdir.glob("run_*\shieldhit*.log")`
See https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob

grzanka (approved):
ok

PR #274, opened 2023-03-10
limit CPU resources for flask worker

### LINKED ISSUE #273: Limit CPU resources for yaptide worker

grzanka:
Up to 2 cores will be left for flask.
This way we partially avoid situation where MC simulation will make flask unresponsive.

To be tested with https://github.com/yaptide/deploy/pull/21

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,10 @@ services:
                 SHIELDHIT_PATH: ${SHIELDHIT_PATH-/shieldhit}
             dockerfile: Dockerfile-worker
         image: yaptide_worker
+        deploy:
+            resources:
+                limits:
+                    cpus: ${MAX_CORES-/1}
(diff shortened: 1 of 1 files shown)


grzanka:
image

grzanka:
image

grzanka:
Seems to work, gives < 100% of load on all cores

p1003:
image
I tested it on smaller machine and load is over but close to 50%. Taking into account, that it had only 4 CPUs, available number for worker was 2 (50%) so it makes sense. LGTM

grzanka:
> image I tested it on smaller machine and load is over but close to 50%. Taking into account, that it had only 4 CPUs, available number for worker was 2 (50%) so it makes sense. LGTM

If looks OK, then give it approve :)

p1003 (approved):

PR #276, opened 2023-03-15
python 3.10 and 3.11 support

grzanka:


grzanka:
Included in https://github.com/yaptide/yaptide/pull/327

PR #280, opened 2023-03-16
Run single task on Slurm

### LINKED ISSUE #243: implement API method which starts simulation on a single task (single node)
This should be based on https://github.com/yaptide/yaptide/pull/238/files

Authentication should be based on private keys instead of passwords. Private keys are stored in the DB. 
Ideally the keys should have short expiration date.

### LINKED ISSUE #250: implement API method which is fetching results from multiple tasks works (merging is needed)
You could apply `convertmc` binary or call it as a python script.
This should handle fetching partial results as well.

The idea is to implement a new JSON converter in pymchelper:
https://github.com/DataMedSci/pymchelper/issues/629

and generate JSON on the cluster, then zip it and send to backend.

Hardcode 2 jobs and plgrid-testing partition (which has limit of max 3 concurrent jobs).

p1003:


yaptide/batch/fabric_test.py
```diff
@@ -1,8 +1,18 @@
...
+from paramiko import Ed25519Key
 
-login = "---------------"
-host = "ares"
-connection = Connection(host=f'{login}@{host}.cyfronet.pl', connect_kwargs={"password": "---------------"})
+import os
+
+from pathlib import Path
+
+ROOT_DIR = os.path.dirname(os.path.realpath(__file__))
```
grzanka: use pathlib

yaptide/batch/fabric_test.py
```diff
@@ -1,8 +1,18 @@
...
+ssh_key_path = Path(ROOT_DIR, "id_ed25519")
```
grzanka: with pathlib this would become:
```suggestion
ssh_key_path = ROOT_DIR / "id_ed25519"
```

yaptide/batch/fabric_test.py
p1003: This is for dummy purpose and will be deleted

grzanka: ok, good !

yaptide/batch/batch_methods.py
```diff
@@ -1,26 +1,85 @@
...
     """Dummy version of submit_job"""
+    with tempfile.TemporaryDirectory() as tmp_dir_path:
+        input_files = write_input_files(json_data, Path(tmp_dir_path))
+        script = SHIELDHIT_BASH.format(
+            beam=input_files["beam.dat"],
+            detect=input_files["detect.dat"],
+            geo=input_files["geo.dat"],
+            mat=input_files["mat.dat"]
+        )
+        ssh_key_path = Path(tmp_dir_path, "id_ed25519")
```
grzanka: do we need to write key to the file ? can't we pass it as a string ?

p1003: That was my initial target and I was trying to find the solution but nothing else worked. I will work on it at the end of the PR because more important things are going to be added.

yaptide/batch/batch_methods.py
```diff
@@ -1,59 +1,173 @@
...
+    array_script = ARRAY_SHIELDHIT_BASH.format(
+        root_dir=job_dir,
+        particle_no=str(1000)
+    )
+    collect_script = COLLECT_BASH.format(
+        root_dir=job_dir
+    )
+
+    con.run(f'echo \'{array_script}\' >> {array_file}')
+    con.run(f'chmod 777 {array_file}')
```
grzanka:
```suggestion
    con.run(f'chmod +x {array_file}')
```

yaptide/batch/batch_methods.py
```diff
@@ -1,59 +1,173 @@
...
+    con.run(f'echo \'{submit_script}\' >> {submit_file}')
+    con.run(f'chmod 777 {submit_file}')
```
grzanka:
```suggestion
    con.run(f'chmod +x {submit_file}')
```

yaptide/batch/batch_methods.py
```diff
@@ -1,59 +1,173 @@
...
+    con.run(f'echo \'{collect_script}\' >> {collect_file}')
+    con.run(f'chmod 777 {collect_file}')
```
grzanka:
```suggestion
    con.run(f'chmod +x {collect_file}')
```

grzanka (approved):
lgtm

PR #285, opened 2023-03-28
New files reading methodology

p1003:


yaptide/celery/tasks.py
```diff
@@ -2,210 +2,197 @@
...
+        yield line
+
+
+def read_file(stats: SimulationStats, filepath: Path, task_id: int):
+    """Monitoring function"""
+    run_match = r"\bPrimary particle no.\s*\d*\s*ETR:\s*\d*\s*hour.*\d*\s*minute.*\d*\s*second.*\b"
+    complete_match = r"\bRun time:\s*\d*\s*hour.*\d*\s*minute.*\d*\s*second.*\b"
+    requested_match = r"\bRequested number of primaries NSTAT"
+
+    while True:
```
grzanka: what if the file will never be created ? i.e. shieldhit is crashing before saving logs ?
maybe add some logic with attempt counter.
we can assume the file is created no longer that 30 seconds from the start.
Therefore add a logic to enter except clause only max 30 times

yaptide/celery/tasks.py
```diff
@@ -2,210 +2,197 @@
...
-        self.update_state(state="PROGRESS", meta={"path": tmp_dir_path, "sim_type": param_dict["sim_type"]})
-        try:
-            is_run_ok = runner_obj.run(settings=settings)
-            if not is_run_ok:
-                raise Exception
-        except Exception:  # skipcq: PYL-W0703
-            logfile = simulation_logfile(path=Path(tmp_dir_path, "run_1", "shieldhit_0001.log"))
-            input_files = simulation_input_files(path=tmp_dir_path)
-            return {"logfile": logfile, "input_files": input_files}
+        logs_list = [Path(tmp_dir_path) / f"run_{1+i}" / f"shieldhit_{1+i:04d}.log" for i in range(ntasks)]
```
grzanka, line 159: why not using glob ?
```suggestion
        logs_list = [Path(tmp_dir_path).glob("run_*/shieldhit_*log")]
```

Later check if you got exactly the same number of logs as expected

yaptide/celery/tasks.py
```diff
@@ -2,210 +2,197 @@
...
-    if "beam.dat" not in raw_input_dict:
-        conv_parser = get_parser_from_str(param_dict["sim_type"])
-        return run_parser(parser=conv_parser, input_data=raw_input_dict, output_dir=output_dir)
-
-    for key, file in raw_input_dict.items():
-        with open(Path(output_dir, key), "w") as writer:
-            writer.write(file)
-    return raw_input_dict
+    def __init__(self, ntasks: int, parent, parent_id: str):
+        self.lock = Lock()
```
grzanka, line 32: why do we need so elaborate data structure ?

yaptide/celery/tasks.py
```diff
@@ -2,210 +2,197 @@
...
-            "input_files": input_files,
-            "end_time": datetime.utcnow(),
-            "job_tasks_status": sh12a_simulation_status(dir_path=tmp_dir_path, sim_ended=True)
-        }
+            try:
+                is_run_ok = runner_obj.run(settings=settings)
+                if not is_run_ok:
+                    raise Exception
+            except Exception:  # skipcq: PYL-W0703
+                logfile = simulation_logfile(path=Path(tmp_dir_path, "run_1", "shieldhit_0001.log"))
```
grzanka: why only first log is being read here ?

grzanka (changes requested):
See comments inline

yaptide/celery/tasks.py
p1003: It won't find anything because files are not created yet -> creation is handled by Pymchelper in 164 line

yaptide/celery/tasks.py
p1003: Do you suggest to return all files?

yaptide/celery/tasks.py
p1003: It's class shared by multiple processes. What is so elaborate?

yaptide/celery/tasks.py
grzanka: Yes, then we could see which ones are failing

grzanka: Also - some of them may be missing, handle that case correctly

p1003: Ok

yaptide/celery/tasks.py
grzanka: You are correct !

yaptide/celery/tasks.py
p1003: Done

yaptide/routes/celery_routes.py
```diff
@@ -124,26 +125,23 @@ def delete(user: UserModel):
...
-        param_dict: dict = schema.load(request.args)
-
         json_data: dict = request.get_json(force=True)
         if not json_data:
             return yaptide_response(message="No JSON in body", code=400)
 
-        job = convert_input_files.delay(param_dict=param_dict, raw_input_dict=json_data)
+        sim_type = SimulationModel.SimType.SHIELDHIT.value if "sim_type" not in json_data or\
+            json_data["sim_type"].upper() == SimulationModel.SimType.SHIELDHIT.value else\
+            SimulationModel.SimType.DUMMY.value
```
hendzeld: Can we use parenthesis instead?
It is advised in PEP 8 style guide: https://peps.python.org/pep-0008/#code-lay-out

yaptide/utils/sim_utils.py
```diff
@@ -0,0 +1,127 @@
+from pathlib import Path
+import sys
+
+from pymchelper.estimator import Estimator
+from pymchelper.axis import MeshAxis
+
+# dirty hack needed to properly handle relative imports in the converter submodule
+sys.path.append("yaptide/converter")
```
hendzeld, line 8: Is this system agnostic?

What will happen on windows, or maybe we doesn't care?

hendzeld (commented):
Have one sty;e suggestion and  one question regarding OS support

yaptide/utils/sim_utils.py
grzanka: backend works only on Linux

yaptide/routes/celery_routes.py
p1003: Good point

yaptide/utils/sim_utils.py
p1003: Furthermore since Python does not understand relative imports this is necessary workaround for the purpose of importing converter submodule. Maybe there is better solution after 2 years but we weren't focused on refactoring this

grzanka (approved):
lgtm

PR #286, opened 2023-03-28
fix error while docker-compose

### LINKED ISSUE #281: Error while executing: docker-compose up / down
While trying to execute `docker-compose up` or `docker-compose down` an error occurs:
ERROR: Error while attempting to convert service.yaptide_worker.deploy.resources.limits.cpus to appropriate type: "/1" is not a valid float

kacper1409:
changed env variable syntax in docker-compose.yml

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,7 +30,7 @@ services:
         deploy:
             resources:
                 limits:
-                    cpus: ${MAX_CORES-/1}
+                    cpus: ${MAX_CORES-1.0}
         container_name: yaptide_worker
         command: celery --app yaptide.celery.worker worker -P threads --loglevel=info
         environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,6 +30,7 @@ services:
         deploy:
             resources:
                 limits:
+                    # the explanation of below syntax: https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html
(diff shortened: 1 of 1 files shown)


grzanka:
Is there an issue related to this PR ?

kacper1409:
> Is there an issue related to this PR ?

Yes, there is #281 . Should I include it in title or description?

grzanka:
> > Is there an issue related to this PR ?
> 
> Yes, there is #281 . Should I include it in title or description?

I've added it here:
image
then merging this PR would fix the issue.

docker-compose.yml
```diff
@@ -30,7 +30,7 @@ services:
         deploy:
             resources:
                 limits:
-                    cpus: ${MAX_CORES-/1}
+                    cpus: ${MAX_CORES-1.0}
```
Bogdan2423, line 34: add a comment explaining how this syntax works (provide a link to a website explaining variable expansion with default values)

Bogdan2423 (changes requested):

docker-compose.yml
kacper1409: I think I found something here: https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

grzanka: Good, then add a comment in the `yml` file

Bogdan2423 (approved):

PR #289, opened 2023-03-31
Shieldhit watcher for slurm

p1003:


yaptide/batch/string_templates.py
```diff
@@ -82,9 +87,104 @@
...
+                    "seconds": int(splitted[6]),
+                },
+                "task_state": "COMPLETED"
+            }
+            print(f"Update for task: {{task_id}} - COMPLETED")
+            return
+
+
+if __name__ == '__main__':
+    signal.signal(signal.SIGUSR1, signal.SIG_IGN)
```
grzanka: why do you need to call `signal` ?

p1003: I wanted to make sure, SIGUSR1 is ignored by watcher because it is important only for running shieldhit binary and we don't want the watcher to do anything about it

p1003: Ah wait I didn't understand the question. This is just registering handler for SIGUSR1 as SIG_IGN

yaptide/batch/string_templates.py
```diff
@@ -71,6 +75,7 @@
 GEO_FILE=$ROOT_DIR/input/geo.dat
 MAT_FILE=$ROOT_DIR/input/mat.dat
 DETECT_FILE=$ROOT_DIR/input/detect.dat
+WATCHER_FILE=$ROOT_DIR/watcher.py
```
grzanka:
```suggestion
WATCHER_SCRIPT=$ROOT_DIR/watcher.py
```

yaptide/batch/string_templates.py
```diff
@@ -82,9 +87,15 @@
     wait # wait for all children, this is important!
 }}
 
+FILE_TO_WATCH=$WORK_DIR/shieldhit_`printf %04d $SLURM_ARRAY_TASK_ID`.log
+srun python3 $WATCHER_FILE --filepath=$FILE_TO_WATCH\\
```
grzanka:
```suggestion
srun python3 $WATCHER_SCRIPT --filepath=$FILE_TO_WATCH\\
```

yaptide/batch/watcher.py
```diff
@@ -0,0 +1,90 @@
+from pathlib import Path
+import re
+import time
+import argparse
+import signal
+
+
+def log_generator(thefile):
+    """Generator function for monitoring purpose"""
```
grzanka:
```suggestion
    """Generator equivalent to `tail -f` Linux command. Yields new lines appended to the end of the file. Main purpose is monitoring of the log files"""
```

yaptide/batch/watcher.py
```diff
@@ -0,0 +1,90 @@
...
+    while True:
+        line = thefile.readline()
+        if not line:
```
grzanka, line 23: I would add a timeout as well. If no new line is appended after some time then the generator will exit. This can be implemented using counter in the `while` loop. By default I would set the timeout to 1 hour (3600 seconds or 3600 passes of the loop). 

The timeout could be added as a parameter to the generator with default value o 3600

It may happen that for some reason the log file may stay permanently open for indefinite time and the generator will be stucked. This would cause the monitoring process to be hanging for ever.

yaptide/batch/watcher.py
```diff
@@ -0,0 +1,90 @@
...
+            time.sleep(1)
+            continue
+        yield line
+
+
+def read_file(filepath: Path, job_id: str, task_id: int):  # skipcq: PYL-W0613
+    """Monitoring function"""
```
grzanka:
```suggestion
    """Monitors log file of certain job/task"""
```

yaptide/batch/watcher.py
```diff
@@ -0,0 +1,90 @@
...
+            up_dict = {  # skipcq: PYL-W0612
+                "run_time": {
+                    "hours": int(splitted[2]),
+                    "minutes": int(splitted[4]),
+                    "seconds": int(splitted[6]),
+                },
+                "task_state": "COMPLETED"
+            }
+            print(f"Update for task: {task_id} - COMPLETED")
+            return
```
grzanka, line 84: Add another return for whole function

grzanka (changes requested):

yaptide/batch/watcher.py
p1003: For what reason? After generator breaks, for loop in `read_file` function will also break.

yaptide/batch/watcher.py
```diff
@@ -0,0 +1,109 @@
...
+import signal
+
+
+RUN_MATCH = r"\bPrimary particle no.\s*\d*\s*ETR:\s*\d*\s*hour.*\d*\s*minute.*\d*\s*second.*\b"
+COMPLETE_MATCH = r"\bRun time:\s*\d*\s*hour.*\d*\s*minute.*\d*\s*second.*\b"
+REQUESTED_MATCH = r"\bRequested number of primaries NSTAT"
+TIMEOUT_MATCH = r"\bTimeout occured"
+
+
+def log_generator(thefile, timeout: int = 3600):
```
grzanka:
```suggestion
def log_generator(thefile, timeout: int = 3600) -> str:
```

yaptide/batch/watcher.py
```diff
@@ -0,0 +1,109 @@
...
+    Main purpose is monitoring of the log files
+    """
+    sleep_counter = 0
+    while True:
+        line = thefile.readline()
+        if not line:
+            time.sleep(1)
+            sleep_counter += 1
+            if sleep_counter >= timeout:
+                yield "Timeout occured"
```
grzanka:
```suggestion
                return "Timeout occured"
```

shouldn't we ensure that the generator is exhausted ?

yaptide/batch/watcher.py
```diff
@@ -0,0 +1,109 @@
...
+            }
+            print(f"Update for task: {task_id} - COMPLETED")
+            return
+
+        elif re.search(TIMEOUT_MATCH, line):
+            up_dict = {  # skipcq: PYL-W0612
+                "task_state": "FAILED"
+            }
+            print(f"Update for task: {task_id} - TIMEOUT")
+            return
```
grzanka, line 91:
```suggestion
            return
    return
```

yaptide/batch/watcher.py
```diff
@@ -0,0 +1,109 @@
+from pathlib import Path
+import re
+import time
+import argparse
+import signal
```
grzanka: sort imports

grzanka:
It seems that watcher script is not properly transmitted:
```
[ares][plgkongruencj@login01 1680539074676812]$ pwd
/net/ascratch/people/plgkongruencj/yaptide_runs/1680539074676812
[ares][plgkongruencj@login01 1680539074676812]$ tail -n 20 watcher.py 
            return
    return


if __name__ == __main__:
    signal.signal(signal.SIGUSR1, signal.SIG_IGN)

    parser = argparse.ArgumentParser()
    parser.add_argument(--filepath, type=str)
    parser.add_argument(--job_id, type=str)
    parser.add_argument(--task_id, type=int)
    args = parser.parse_args()
    filepath_arg = Path(args.filepath)
    job_id_arg = args.job_id
    task_id_arg = args.task_id

    print(filepath=filepath_arg, job_id=job_id_arg, task_id=task_id_arg)

    read_file(filepath=filepath_arg, job_id=job_id_arg, task_id=task_id_arg)
```

I see following error:
```
[ares][plgkongruencj@login01 1680539074676812]$ cat slurm-2169041_1.out 
/net/ascratch/people/plgkongruencj/yaptide_runs/1680539074676812/workspaces/task_0001
/net/ascratch/people/plgkongruencj/yaptide_runs/1680539074676812/workspaces/task_0001
Traceback (most recent call last):
  File "/net/ascratch/people/plgkongruencj/yaptide_runs/1680539074676812/watcher.py", line 95, in <module>
    if __name__ == __main__:
NameError: name '__main__' is not defined
srun: error: ac0782: task 0: Exited with exit code 1
```

There are missing `"` characters around `__main__`

yaptide/batch/string_templates.py
```diff
@@ -27,6 +28,9 @@
 cat << EOF > $DETECT_FILE
 {detect}
 EOF
+cat << EOF > $WATCHER_SCRIPT
+{watcher}
+EOF
```
grzanka, line 33: I am afraid using cat may string double qoutes and other unexpected consequences.
See for example:
https://unix.stackexchange.com/questions/532036/how-to-display-double-quotes

The same apply to other files.

Why not copy these files using fabric ?

grzanka:
Also: simulation will fail if more input file are available (for example sobp.dat):

```

/net/ascratch/people/plgkongruencj/yaptide_runs/1680539732718378/workspaces/task_0001
/net/ascratch/people/plgkongruencj/yaptide_runs/1680539732718378/workspaces/task_0001
  
  *** SHIELD-HIT12A-beta *** Copyright (c) 2023
  Institute of Nuclear Research RAS, Moscow, Russia and
  Dept. of Physics and Astronomy, Aarhus University, Aarhus, Denmark and
  Dept. of Experimental Clinical Oncology Aarhus University Hospital, Aarhus, Denmark.
  
  Version: v1.0.0
  Build date: Mon, 03 Apr 2023 18:34:22 +0200  
  
  See /net/pr2/projects/plgrid/plggccbmc/shieldhit --help for command line options.
  
  Working directory: /net/ascratch/people/plgkongruencj/yaptide_runs/1680539732718378/workspaces/task_0001/

*** Error: sh_fopen() cannot open file /net/ascratch/people/plgkongruencj/yaptide_runs/1680539732718378/workspaces/task_0001//sobp.dat for reading.

```

yaptide/batch/string_templates.py
p1003: I wanted to minimize the number of commands running via ssh.
It worked for `.dat` files without any problems

grzanka: Yes, but what if we introduce `"` in the dat files ?
Also - some of the input files, like `sobp.dat` can be pretty large (~1MB). Putting that large files into strings is not a good idea.

What about on-the-fly compression of all what is needed to run SH12A into a single compressed file ? For example:
https://docs.python.org/3/library/zipfile.html

On the server side we can assume that `zip` is always available.

grzanka (approved):
ok, comments will be moved to a separate issues

PR #487, opened 2023-06-28
fix for requested primaries and simulated primaries are not equal in completed task

### LINKED ISSUE #484: Requested primaries and simulated primaries are not equal in COMPLETED task
image
Frontend handles it for now and logs an error, but the backend response must provide accurate info.

grzanka:


yaptide/celery/utils/utils.py
```diff
@@ -144,26 +149,36 @@ def read_file(filepath: Path, simulation_id: int, task_id: str, update_key: str)
...
+                send_task_update(simulation_id, task_id, update_key, up_dict)
+                update_time = utc_now.timestamp()
+            # block of code below makes the updates not so frequent
+            # updates comes every `seconds_before_next_update` seconds
+            # this is convenient for long simulations, but we may also miss this way some updates,
+            # like the one when simulation is finished (simulated_primaries == requested_primaries)
+            # therefore we send update when simulation is finished
+            if simulated_primaries == requested_primaries:
+                send_task_update(simulation_id, task_id, update_key, up_dict)
+                update_time = utc_now.timestamp()
```
p1003, line 174: Blocks of code after both if statements are equal. Furthermore it is possible for both statements to be true, so task will be updated 2 times in the same iteration. Suggestions:
- merge statements with `or`
- or at the end of first if add `continue`

tests/integration/conftest.py
```diff
@@ -10,6 +11,42 @@
...
+    payload_dict["ntasks"] = 2
+    payload_dict["input_json"]["beam"]["numberOfParticles"] = 12
+
+    # reduce number of segments to get results which are easier to print
+    for detector in payload_dict["input_json"]["detectorManager"]["detectors"]:
+        for segments_type in ["xSegments", "ySegments", "zSegments"]:
+            # if segments_type is missing (for example xSegments is missing for cylinder detector)
+            # then we skip it by comparing default value 2 > 1
+            if detector["geometryData"]["parameters"].get(segments_type, 2) > 1:
+                detector["geometryData"]["parameters"][segments_type] = 4
```
hendzeld, line 29: I am not sure if I understand this correctly it seems that we are setting the value to 4 if:

- detector["geometryData"]["parameters"][segments_type] is not defined
- detector["geometryData"]["parameters"][segments_type] is ~less than  or equal~ greater than 1

To update only existing values we can use
```
if segments_type in detector["geometryData"]["parameters"]
```

grzanka: I would like to update that value if one of following condition holds:
 - `segments_type`  key doesn't exists
 - `segments_type`  key does exists and corresponding value is greater than one

hendzeld: Right, 
Ok. So the logic is correct, I was just misguided by the comment above:
```
# then we skip it by comparing default value 2 > 1
```

Feel free to close this discussion if I am missing some context and this comment is a valid one.

grzanka: I could rephrase the comment, in fact the code is a bit tricky. Is segment key is missing, then a default value 2 is retrieved. Then a comparison 2 > 1 is evaluated which is true.

grzanka:
Closing in favour of https://github.com/yaptide/yaptide/pull/575

Issue #485, opened 2023-06-27
Improve celery job execution


grzanka:
Single celery job should correspond to single shieldhit process. This would enable better cancelling and better handling of multiple celery instances. Merging of results could happen as separate celery job

p1003:
@grzanka consider closing issue this due to the merge of https://github.com/yaptide/yaptide/pull/492

grzanka:
Fixed with #492

PR #492, opened 2023-07-03
Rework Celery workflow

### LINKED ISSUE #515: Once collect job is finished it should upload the results to the backend
Currently the results are being copied using SFTP, instead we should use the POST endpoint (/results ?) allowing to store the results in the DB. 
Tricky part is that the submit job should know the secret allowing it to use the POST endpoint.

We could use watcher.py script , the best would be to use standard python library, avoiding `requests` library.
This would make the solution more portable.
```
import http.client
```

p1003:
Related to https://github.com/yaptide/yaptide/issues/529 and https://github.com/yaptide/yaptide/issues/485
Fixes https://github.com/yaptide/yaptide/issues/515


yaptide/batch/watcher.py
```diff
@@ -4,21 +4,25 @@
...
+
 
 RUN_MATCH = r"\bPrimary particle no.\s*\d*\s*ETR:\s*\d*\s*hour.*\d*\s*minute.*\d*\s*second.*\b"
 COMPLETE_MATCH = r"\bRun time:\s*\d*\s*hour.*\d*\s*minute.*\d*\s*second.*\b"
 REQUESTED_MATCH = r"\bRequested number of primaries NSTAT"
 TIMEOUT_MATCH = r"\bTimeout occured"
 
 
-def log_generator(thefile, timeout: int = 3600) -> str:
+def log_generator(thefile: TextIOWrapper, timeout: int = 3600):
```
grzanka:
```suggestion
def log_generator(thefile: TextIOWrapper, timeout: int = 3600) -> str:
```

p1003:
Working with UI at current state
image
image

grzanka (approved):

PR #511, opened 2023-07-24
472 topas s3 download

Bogdan2423:
For now the `install --name topas` command downloads from S3 and extracts the correct version of TOPAS.
I assume the TOPAS files are kept in S3 bucket as an uncompressed tar file with the tag "version". Geant files are kept in another bucket also as uncompressed tar files, each with the tag "topas_versions", containing information about supported TOPAS versions.
The following env variables need to be set to enable the command to download TOPAS: S3_ENDPOINT, S3_ACCESS_KEY, S3_SECRET_KEY, S3_TOPAS_BUCKET, S3_TOPAS_KEY, S3_GEANT_BUCKET, S3_TOPAS_VERSION

### DIFF
--- a/tests/test_download_topas.py
+++ b/tests/test_download_topas.py
@@ -1,0 +1,36 @@
+import os
+from pathlib import Path
+from yaptide.admin.simulators import download_topas_from_s3
+
+
+def check_if_environment_variables_set() -> bool:
+    """Check if environment variables are set"""
+    result = True
+    if not 'S3_ENDPOINT' in os.environ:
+        result = False
+    if not 'S3_ACCESS_KEY' in os.environ:
+        result = False
+    if not 'S3_SECRET_KEY' in os.environ:
+        result = False
+    if not 'S3_TOPAS_BUCKET' in os.environ:
+        result = False
+    if not 'S3_TOPAS_KEY' in os.environ:
+        result = False
+    if not 'S3_GEANT_BUCKET' in os.environ:
+        result = False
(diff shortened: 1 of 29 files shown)


tests/test_download_topas.py
```diff
@@ -0,0 +1,36 @@
...
+    if not 'S3_GEANT_BUCKET' in os.environ:
+        result = False
+    if not 'S3_TOPAS_VERSION' in os.environ:
+        result = False
+    return result
+
+def test_if_topas_downloaded(tmpdir):
+    """Check if shieldhit is downloaded and can be executed"""
+    if check_if_environment_variables_set():
+        bucket = os.getenv("S3_SHIELDHIT_BUCKET")
```
grzanka, line 28: that should be TOPAS, not shieldhit

tests/test_download_topas.py
```diff
@@ -0,0 +1,36 @@
...
+        result = False
+    if not 'S3_TOPAS_KEY' in os.environ:
+        result = False
+    if not 'S3_GEANT_BUCKET' in os.environ:
+        result = False
+    if not 'S3_TOPAS_VERSION' in os.environ:
+        result = False
+    return result
+
+def test_if_topas_downloaded(tmpdir):
```
grzanka, line 25: that test should be skipped on Windows as we are not planning to use TOPAS windows builds

grzanka:
Closing in favor of https://github.com/yaptide/yaptide/pull/536

PR #526, opened 2023-08-05
Slurm monitoring

### LINKED ISSUE #524: Notify backend about status of SLURM jobs
Watcher SLURM script should notify backend about the state (pending/running/completed) and the progress of slurm jobs.
See https://github.com/yaptide/yaptide/issues/515 as well

p1003:


yaptide/routes/batch_routes.py
```diff
@@ -135,6 +139,8 @@ def get(user: KeycloakUserModel):
 
         job_tasks_status = [task.get_status_dict() for task in tasks]
 
+        simulation
+
```
grzanka:
```suggestion
```

yaptide/celery/utils/utils.py
```diff
@@ -81,7 +81,7 @@ def send_task_update(simulation_id: int, task_id: str, update_key: str, update_d
         "update_key": update_key,
         "update_dict": update_dict
     }
-    res: requests.Response = requests.Session().post(url=f"{flask_url}/tasks/update", json=dict_to_send)
+    res: requests.Response = requests.Session().post(url=f"{flask_url}/tasks", json=dict_to_send)
```
grzanka, line 84: why was this changed ?

p1003: Because `tasks` endpoint did not exist and it is bad practise to have child endpoint when there is no parent endpoint. The problem occures in some other places currently.

yaptide/batch/watcher.py
```diff
@@ -2,13 +2,19 @@
...
+from urllib import request
+import logging
+from datetime import datetime
 
 
 RUN_MATCH = r"\bPrimary particle no.\s*\d*\s*ETR:\s*\d*\s*hour.*\d*\s*minute.*\d*\s*second.*\b"
 COMPLETE_MATCH = r"\bRun time:\s*\d*\s*hour.*\d*\s*minute.*\d*\s*second.*\b"
 REQUESTED_MATCH = r"\bRequested number of primaries NSTAT"
 TIMEOUT_MATCH = r"\bTimeout occured"
+HARDCODED_BACKEND_URL = "https://yap-dev.c3.plgrid.pl:8443"
```
grzanka: Can you send that URL when submitting job ? On the other hand backend itself could retrieve its own URL from an environmental variable. I can extend the deploy script to handle it.

yaptide/batch/watcher.py
```diff
@@ -30,9 +36,37 @@ def log_generator(thefile, timeout: int = 3600) -> str:
         yield line
 
 
-def read_file(filepath: Path, job_id: str, task_id: int):  # skipcq: PYL-W0613
+def send_task_update(sim_id: int, task_id: str, update_key: str, update_dict: dict) -> bool:
```
grzanka: I'd add an URL of backend as a parameter to this method. This would simplify testing.

yaptide/batch/watcher.py
```diff
@@ -97,13 +135,15 @@ def read_file(filepath: Path, job_id: str, task_id: int):  # skipcq: PYL-W0613
 
     parser = argparse.ArgumentParser()
     parser.add_argument("--filepath", type=str)
-    parser.add_argument("--job_id", type=str)
+    parser.add_argument("--sim_id", type=int)
     parser.add_argument("--task_id", type=int)
+    parser.add_argument("--update_key", type=str)
```
grzanka, line 155: Here we could add URL as well.

yaptide/persistence/models.py
```diff
@@ -189,7 +189,7 @@ class TaskModel(db.Model):
     # currently for Celery one simulation is one job, but in future we might want to split it into multiple tasks
     # example celery job_ids are: 1a2b3c4d-5e6f-7g8h-9i0j-1k2l3m4n5o6p
     # example SLURM job_ids are: 12345678_12
-    task_id: Column[str] = db.Column(db.String, nullable=False, unique=True, doc="Task ID")
+    task_id: Column[str] = db.Column(db.String, nullable=False, doc="Task ID")
```
grzanka, line 192: Why this is not unique anymore ?

yaptide/routes/batch_routes.py
```diff
@@ -79,14 +79,15 @@ def post(user: KeycloakUserModel):
         input_dict_to_save["number_of_all_primaries"] = number_of_all_primaries
         input_dict_to_save["input_files"] = files_dict
 
-        result = submit_job(payload_dict=payload_dict, files_dict=files_dict, user=user, cluster=cluster)
+        result = submit_job(payload_dict=payload_dict, files_dict=files_dict, user=user,
```
grzanka, line 82: At this point I'd extract backend URL from an environment variable. If not set I'd say it should go to localhost. Then this should go into the submit script.

grzanka (changes requested):
See comments

yaptide/batch/watcher.py
p1003: Of course I can. This was temporary HARDCODED solution. When you change deploy I can change it here

yaptide/persistence/models.py
p1003: Because task id is unique only within single simulation and that is how it's uniqueness was redefined here: ```
    __table_args__ = (
        UniqueConstraint('simulation_id', 'task_id', name='_simulation_id_task_id_uc'),
    )
    ```

grzanka: Good !

yaptide/batch/watcher.py
grzanka: Start changing it here, I will adjust deploy later. You can always test it by `MYNEWENV=myblablavalu docker compose up` and adding a line like that:
https://github.com/yaptide/yaptide/blob/28344fbd0ea4aa79fa5fa637cbdafa3b95f1ebe1/docker-compose.yml#L18
and like that:
https://github.com/yaptide/yaptide/blob/28344fbd0ea4aa79fa5fa637cbdafa3b95f1ebe1/Dockerfile-flask#L9

I'd call this name BACKEND_EXTERNAL_URL

yaptide/routes/batch_routes.py
p1003: Added in lowest possible point

grzanka (approved):

PR #638, opened 2023-10-08
288 proper logging

### LINKED ISSUE #288: proper logging
Add more logging messages on a debug level. We should be able to understand more about timing of application by looking on the logs.
Logs verbosity level can be easily controlled by celery or flask switches

hendzeld:
## Changes

Adds logging for:
https://github.com/yaptide/yaptide/blob/de3e55e99746f00518d8538f4712ff3bb6bd602c/yaptide/batch/watcher.py#L72

## Resolves 
#288

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -41,7 +41,7 @@ services:
             - CELERY_BROKER_URL=redis://redis:6379/0
             - CELERY_RESULT_BACKEND=redis://redis:6379/0
             - BACKEND_INTERNAL_URL=http://yaptide_flask:6000
-            - LOG_LEVEL_CELERY=INFO
+            - LOG_LEVEL_ROOT=${LOG_LEVEL_ROOT:-INFO}
         depends_on:
             - redis
             - yaptide_flask
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -41,6 +41,7 @@ services:
             - CELERY_BROKER_URL=redis://redis:6379/0
             - CELERY_RESULT_BACKEND=redis://redis:6379/0
             - BACKEND_INTERNAL_URL=http://yaptide_flask:6000
+            - LOG_LEVEL_CELERY=INFO
         depends_on:
             - redis
             - yaptide_flask
--- a/run_worker.sh
+++ b/run_worker.sh
@@ -22,6 +22,4 @@ if [ -z "$S3_TOPAS_BUCKET" ] || [ -z "$S3_TOPAS_KEY" ] || [ -z "$S3_TOPAS_VERSIO
         ./yaptide/admin/simulators.py install --name topas --path /simulators
 fi
 
-export LOG_LEVEL_CELERY=${LOG_LEVEL_CELERY-INFO}
-
-celery --app yaptide.celery.worker worker -E --loglevel="$LOG_LEVEL_CELERY" -P eventlet --hostname yaptide-worker
+celery --app yaptide.celery.worker worker -E --loglevel="$LOG_LEVEL_ROOT" -P eventlet --hostname yaptide-worker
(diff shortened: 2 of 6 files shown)


docker-compose.yml
```diff
@@ -39,6 +39,7 @@ services:
             - CELERY_BROKER_URL=redis://redis:6379/0
             - CELERY_RESULT_BACKEND=redis://redis:6379/0
             - BACKEND_INTERNAL_URL=http://yaptide_flask:6000
+            - LOG_LEVEL_CELERY=INFO
```
grzanka:
```suggestion
            - LOG_LEVEL_CELERY=${LOG_LEVEL_CELERY-"INFO"}
```

I would use default value and some control from env variable

grzanka: If we set enviroment variable outside docker compose, i.e. in .env file it wont be propagated here

run_worker.sh
```diff
@@ -22,4 +22,6 @@ if [ -z "$S3_TOPAS_BUCKET" ] || [ -z "$S3_TOPAS_KEY" ] || [ -z "$S3_TOPAS_VERSIO
         ./yaptide/admin/simulators.py install --name topas --path /simulators
 fi
 
-celery --app yaptide.celery.worker worker -E --loglevel=info -P eventlet --hostname yaptide-worker
\ No newline at end of file
+export LOG_LEVEL_CELERY=${LOG_LEVEL_CELERY-INFO}
```
grzanka: That won't be necessary if we adjust docker compose

grzanka:
This seems to work fine, can you adjust it to latest changes ?

image

grzanka:
This PR may affect your work:
https://github.com/yaptide/yaptide/pull/650/files

image

docker-compose.yml
hendzeld: Ok, I will add the default value then.

I modified the name of this variable because it actually changes the logging level of `root logger`. 
Not only `celery.*` (tasks etc)

obraz

Is the new name ok, or maybe the previous one was better?

tests/integration/test_celery.py
```diff
@@ -27,14 +27,15 @@ def test_celery_run_simulation(celery_app,
                                client,
                                add_simulators_to_path_variable,
                                modify_tmpdir,
-                               shieldhit_binary_installed):
+                               shieldhit_binary_installed, monkeypatch):
```
grzanka: take a look here:
https://github.com/yaptide/yaptide/blob/288-proper-logging/tests/integration/conftest.py#L68

Are you sure that `monkeypatch.setenv` modifies only enviroment in this particular function and not leaving modified state in the subsequent ones ?

tests/integration/test_celery.py
```diff
@@ -27,14 +27,15 @@ def test_celery_run_simulation(celery_app,
...
     """Test run_simulation task with SHIELD-HIT12A binary
     Current Windows demo version version of SHIELD-HIT12A has a bug, so it cannot parse more elaborated input files.
     Parser relies on rewind function, which does not work properly on Windows, see:
     https://stackoverflow.com/questions/47256223/why-does-fseek-0-seek-cur-fail-on-windows/47256758#47256758
     So to bypass this issue we restrict the detect configuration to only one output and no filter.
     Below goes the code which reduces the detect.dat.
     """
+    monkeypatch.setenv("SKIP_CELERY_MONKEY_PATCHING", "1")
```
grzanka:
```suggestion
```

tests/integration/test_celery.py
grzanka:
```suggestion
                               shieldhit_binary_installed):
```

grzanka (approved):

PR #650, opened 2023-10-11
adjust dockerfile and script for worker

### LINKED ISSUE #637: adjust dockerfiles machinery

Bogdan2423:


Dockerfile-worker
```diff
@@ -1,8 +1,18 @@
...
+    libexpat1-dev \
+    libgl1-mesa-dev \
+    libglu1-mesa-dev \
+    libxt-dev \
+    xorg-dev \
+    build-essential \
+    libharfbuzz-dev \
+    gfortran \
+    && \
+    rm -rf /var/lib/apt/lists/*
```
grzanka: what is the impact of installation of these packages on the docker image size ?

Dockerfile-worker
```diff
@@ -1,8 +1,18 @@
 FROM python:3.10-slim
 
-# install gfortran via apt and clean cache
+# install libraries required to run simulators via apt and clean cache
 # skipcq: DOK-DL3008
-RUN apt-get update && apt-get install -y --no-install-recommends gfortran && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    libexpat1-dev \
```
grzanka: Can we make installation of TOPAS optional. It can be driven by presence of necessary env variables. If the variables are set then the necessary apt packages are being installed. This could either be a smart one-liner or the installation of TOPAS requirements can be moved to a bash script.

This way we will have smaller container image and faster deploy

Dockerfile-worker
Bogdan2423: image size with these packages (but without TOPAS): 20.3MB (virtual 732MB)
vs size without these packages: 483MB (virtual 1.19GB)

Dockerfile-worker
Bogdan2423: I made it so if any of TOPAS variables are not set then the libraries are not downloaded

.env
```diff
@@ -0,0 +1,7 @@
+S3_ENDPOINT=https://s3.cloud.cyfronet.pl
+S3_ACCESS_KEY=plgs3n-plggyaptide-plgmatwro
+S3_SECRET_KEY=B1uR3DZSYzgC8gQAEQi956ZPuJAe87oX1XqYSO4C
+S3_TOPAS_BUCKET=topas
+S3_TOPAS_KEY=topas.tar
+S3_GEANT_BUCKET=geant
+S3_TOPAS_VERSION=3.9
```
grzanka: Your secrets have leaked, please regenerate your credentials as soon as possible !!

.env
```diff
@@ -0,0 +1,7 @@
+S3_ENDPOINT=https://s3.cloud.cyfronet.pl
+S3_ACCESS_KEY=plgs3n-plggyaptide-plgmatwro
+S3_SECRET_KEY=B1uR3DZSYzgC8gQAEQi956ZPuJAe87oX1XqYSO4C
```
grzanka:
```suggestion
```

.env
Bogdan2423: done

Dockerfile-worker
```diff
@@ -17,6 +17,9 @@ ENV PATH="/usr/local/app/bin:${PATH}"
 RUN chgrp -R 0 /usr/local/app && \
     chmod -R g=u /usr/local/app
 
+# set variable with path to Geant4 files required by TOPAS simulator
+ENV TOPAS_G4_DATA_DIR="/usr/local/app/bin/geant"
```
grzanka: Why do you locate data files in the `bin` directory ? This directory should be for binaries only.
Also - the binaries are pretty large, having very large docker images is not optimal.
Maybe better practice would be to create another docker volume, called "simulators_data" and mount it somewhere in the dockerfile worker ?

There is already a volume dedicated to storage of application between restarts, see:
https://github.com/yaptide/yaptide/blob/532ed734bfc17bf4abc1ad0afa1f27d19a8c8c08/docker-compose.yml#L11

I would avoid using this particular storage for keeping simulators data.

What about a volume called `simulators` and mounting it in `/simulators` ?
Then it may have:
  - `/simulators/geant4/bin`
  - `/simulators/geant4/data`
  - `/simulators/shieldhit12a/bin`

All `bin` directories could be added then to the PATH.

I would also avoid using `geant` and use `geant4`

Bogdan2423: Is something like this good?
obraz

yaptide/admin/simulators.py
```diff
@@ -43,7 +43,9 @@ class SimulatorType(IntEnum):
 topas_key = os.getenv('S3_TOPAS_KEY')
 topas_version = os.getenv('S3_TOPAS_VERSION')
 geant_bucket_name = os.getenv('S3_GEANT_BUCKET')
-installation_path = Path(__file__).resolve().parent.parent.parent / 'bin'
+shieldhit_path = Path('/simulators') / 'shieldhit12a' / 'bin'
```
grzanka: I would extract installation path to a click parameter. Hardcoding it to `/simulators` would make it impossible to work with such code locally.

Dockerfile-worker
grzanka: good

.env
grzanka, line 1: this file must be present in the repo but should be empty

Dockerfile-worker
```diff
@@ -11,12 +11,16 @@ ENV CELERYD_TIME_LIMIT=36000
...
+ENV PATH="/simulators/shieldhit12a/bin:${PATH}"
+ENV PATH="/simulators/topas/bin:${PATH}"
 
 # openshift permissions
 RUN chgrp -R 0 /usr/local/app && \
     chmod -R g=u /usr/local/app
 
+# set variable with path to Geant4 files required by TOPAS simulator
+ENV TOPAS_G4_DATA_DIR="/simulators/geant4"
+
```
grzanka: can you move that above openshift stuff ? we should gather simulator related lines close to each other

run_worker.sh
```diff
@@ -1,5 +1,25 @@
...
+
+        apt-get install -y --no-install-recommends \
+        libexpat1-dev \
+        libgl1-mesa-dev \
+        libglu1-mesa-dev \
+        libxt-dev \
+        xorg-dev \
+        build-essential \
+        libharfbuzz-dev \
+        gfortran \
```
grzanka:
```suggestion
        gfortran
```

.env
grzanka: image
image

the file is still removed. you should add it back

yaptide/admin/simulators.py
```diff
@@ -317,23 +318,16 @@ def download_topas_from_s3(bucket: str = topas_bucket_name,
...
+    topas_file_contents.extractall(path=path)
+    topas_extracted_path = path / "topas" / "bin" / "topas"
     topas_extracted_path.chmod(0o700)
-    logging.info("Installed TOPAS into %s", topas_file_path)
-    click.echo(f"Installed TOPAS into {topas_file_path}")
+    logging.info("Installed TOPAS into %s", path)
+    click.echo(f"Installed TOPAS into {path}")
 
-    geant_files_path = path / "geant"
+    geant_files_path = path / "geant4"
```
grzanka: can you rename to `geant4_files_path` ?

yaptide/admin/simulators.py
```diff
@@ -423,6 +417,7 @@ def derive_key(encryption_password: str = password, encryption_salt: str = salt)
 
 @run.command
 @click.option('--name', type=click.Choice([sim.name for sim in SimulatorType]))
+@click.option('--path')
```
grzanka: Consider Path type from https://click.palletsprojects.com/en/8.1.x/parameters/#parameter-types

image

tests/test_download_topas.py
```diff
@@ -25,8 +25,8 @@ def test_if_topas_downloaded(tmpdir):
         version = os.getenv("S3_TOPAS_VERSION")
         geant_bucket = os.getenv("S3_GEANT_BUCKET")
         assert download_topas_from_s3(bucket=bucket, key=key, version=version, geant_bucket=geant_bucket, path=tmpdir) is True
-        expected_path = Path(tmpdir / "topas" / "topas" / "bin" / "topas")
-        expected_geant_path = Path(tmpdir / "geant")
+        expected_path = Path(tmpdir / "topas" / "bin" / "topas")
+        expected_geant_path = Path(tmpdir / "geant4_files_path")
         assert expected_path.exists(), "Expected TOPAS path does not exist."
         assert expected_path.stat().st_size > 0, "Expected TOPAS path is empty."
         assert expected_geant_path.exists(), "Expected Geant path does not exist."
```
grzanka:
```suggestion
        assert expected_path.exists(), "Expected TOPAS path does not exist."
        assert expected_path.stat().st_size > 0, "Expected TOPAS path is empty."
        expected_geant4_path = Path(tmpdir / "geant4_files_path")
        assert expected_geant4_path .exists(), "Expected Geant4 path does not exist."
```

yaptide/admin/simulators.py
```diff
@@ -423,6 +417,7 @@ def derive_key(encryption_password: str = password, encryption_salt: str = salt)
 
 @run.command
 @click.option('--name', type=click.Choice([sim.name for sim in SimulatorType]))
+@click.option('--path', type=click.Path(file_okay=False))
```
grzanka:
```suggestion
@click.option('--path', type=click.Path(file_okay=False, path_type=Path))
```

you can ask click to make conversion to path

grzanka (approved):

Issue #532, opened 2023-08-07
Reduce time for job submission in POST request


grzanka:
# The problem

The `jobs/batch` POST request https://github.com/yaptide/yaptide/blob/c420a4172295f07b342d29763be6caa9f8b28bec/yaptide/routes/batch_routes.py#L22 to submit the batch jobs takes about 8 seconds. Most of the time is being spent on talking to SLURM via SSH:
https://github.com/yaptide/yaptide/blob/c420a4172295f07b342d29763be6caa9f8b28bec/yaptide/batch/batch_methods.py#L42

This means that UI is freezed for that time, here are the timings:
![image](https://github.com/yaptide/yaptide/assets/7374733/61d33719-33ba-420c-8e27-54afcdfff576)
![image](https://github.com/yaptide/yaptide/assets/7374733/5f6e1a9c-f67f-43b7-aa5c-584786d7504c)

We should come up with some strategy to reduce that time.

# Solution 1. Using existing libraries

The simplest (but ugly hack) would be to use some asynchronous call, like this example written by ChatGPT.
This relies on introducing another endpoint which checks if the job was submitted.

```python
from flask import Flask, jsonify, request
import asyncio
from sqlalchemy import create_engine, Column, Integer, String, DateTime
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy.orm import sessionmaker
import datetime

app = Flask(__name__)

# Initialize SQLAlchemy with SQLite database
DATABASE_URL = 'sqlite:///job_results.db'  # SQLite database file
engine = create_engine(DATABASE_URL)
Base = declarative_base()
Session = sessionmaker(bind=engine)
session = Session()


# Model for storing job results in the database
class JobResult(Base):
    __tablename__ = 'job_results'

    id = Column(Integer, primary_key=True)
    job_id = Column(String, unique=True)
    status = Column(String)
    result = Column(String)
    created_at = Column(DateTime, default=datetime.datetime.utcnow)

    def __init__(self, job_id, status, result=None):
        self.job_id = job_id
        self.status = status
        self.result = result


async def submit_the_slurm_job(data):
    try:
        # Your slurm job submission logic here
        # ...

        # Simulate a time-consuming job submission
        await asyncio.sleep(5)

        # Store the job result in the database
        job_result = JobResult(job_id=asyncio.current_task().get_name(), status="Job completed successfully")
        session.add(job_result)
        session.commit()

        return "Job completed successfully"

    except asyncio.CancelledError:
[...]

PR #656, opened 2023-10-12
Introduction of UNKNOWN state

### LINKED ISSUE #647: Introduce  "UNKNOWN" job state as described here:
@p1003 what about the "UNKNOWN" state as described here:

> In case the notification system (watcher.py) is broken and in the database we have no information at all, then all the tasks and simulations in such state should have a new state, called UNKNOWN. This may be implemented easily on creation of the objects in the DB, instead of creating the object in PENDING state, give them state UNKNOWN.
> 
> For Celery transition from UNKNOWN to PENDING happens when simulation will get job_id from Celery.
> For SLURM this transition happens when we obtain job id from slurm. Here it may be time consuming as to get job id we need to call slurm via fabric/ssh.
> Simulations with UNKNOWN state will have empty list of tasks. This should be properly displayed in the UI.
> 
> When listing simulations we should list simulation with UNKNOWN state as well.

_Originally posted by @grzanka in https://github.com/yaptide/yaptide/issues/529#issuecomment-1743048294_

p1003:


grzanka:
Those are the states:
```
class EntityState(Enum):
    """Job state types - move it to more utils like place in future"""

    UNKNOWN = "UNKNOWN"
    PENDING = "PENDING"
    RUNNING = "RUNNING"
    CANCELED = "CANCELED"
    COMPLETED = "COMPLETED"
    FAILED = "FAILED"
```


Is this a correct state graph ?

```mermaid
---
title: Job states
---
stateDiagram-v2
    [*] --> UNKNOWN

    UNKNOWN --> PENDING
    PENDING --> RUNNING
    RUNNING --> COMPLETED
    COMPLETED --> [*]

    RUNNING --> FAILED
    FAILED --> [*]

    PENDING --> CANCELED
    RUNNING --> CANCELED
    CANCELED --> [*]

    UNKNOWN --> CANCELED

```

grzanka:
Is it still draft ?

grzanka:
@p1003 can you discuss with @Derstilon and @ostatni5 how this could be reflected in UI ?

p1003:
@grzanka diagram looks fine but canceling job with `UNKNOWN` state might cause some problems for now. As a quick fix I can forbid canceling jobs in this state, because it might cause some problems:
- if job has `UNKNOWN` state, that might mean the submition has not yet been finished
- Celery tasks need to send their internal ids to database first, because otherwise they cannot be canceled - fix for this might include checking if task is canceled after it sends it's update and sending appropriate message to it, so that he can cancel itself

Currently server sends responses on POST methods submitting jobs, after the simulations' states are changed to `PENDING`. This might change when we start working on this issue [https://github.com/yaptide/yaptide/issues/532](url).

grzanka:
> @grzanka diagram looks fine but canceling job with `UNKNOWN` state might cause some problems for now. As a quick fix I can forbid canceling jobs in this state, because it might cause some problems:
> - if job has `UNKNOWN` state, that might mean the submition has not yet been finished
> - Celery tasks need to send their internal ids to database first, because otherwise they cannot be canceled - fix for this might include checking if task is canceled after it sends it's update and sending appropriate message to it, so that he can cancel itself
> 
> Currently server sends responses on POST methods submitting jobs, after the simulations' states are changed to `PENDING`. This might change when we start working on this issue [https://github.com/yaptide/yaptide/issues/532](url).

Ok, then suggest updated diagram.

grzanka:
Can you add as well a single markdown file in docs folder with state documentation ?

p1003:
```mermaid
---
title: Job states
---
stateDiagram-v2
    [*] --> UNKNOWN

    UNKNOWN --> PENDING
    PENDING --> RUNNING
    RUNNING --> COMPLETED
    COMPLETED --> [*]

    RUNNING --> FAILED
    FAILED --> [*]

    PENDING --> CANCELED
    RUNNING --> CANCELED
    CANCELED --> [*]

```

yaptide/utils/enums.py
```diff
@@ -11,6 +11,7 @@ class PlatformType(Enum):
 class EntityState(Enum):
     """Job state types - move it to more utils like place in future"""
```
grzanka:
```suggestion
    """Job state types"""
```

yaptide/routes/utils/utils.py
```diff
@@ -11,3 +15,32 @@ def check_if_job_is_owned_and_exist(job_id: str, user: UserModel) -> tuple[bool,
     if simulation.user_id == user.id:
         return True, "", 200
     return False, 'Job with provided ID does not belong to the user', 403
+
+
+def determine_input_type(payload_dict: dict) -> Union[str, None]:
```
grzanka:
```suggestion
def determine_input_type(payload_dict: dict) -> Optional[str]:
```

yaptide/routes/utils/utils.py
```diff
@@ -1,5 +1,9 @@
+from typing import Union
```
grzanka:
```suggestion
from typing import Optional
```

grzanka (approved):

PR #664, opened 2023-10-17
Deploy Fluka binary to backend from S3 object

### LINKED ISSUE #394: Deploy FLUKA binary to backend from S3 object
We need to pack fluka binary (rfluka) and all data files (probably hundreds of MBs) and other stuff.

- [ ] #398
- [x] #397
- [ ] #396

### LINKED ISSUE #628: Make yaptide/admin/db_manage.py an executable script
Right now we have pretty long command:

```
docker exec -w /usr/local/app/ yaptide_flask python3 yaptide/admin/db_manage.py add-user admin --password mysecretpassword
```

This could be shortened to:
```
docker exec yaptide_flask yaptide/admin/db_manage.py add-user admin --password mysecretpassword
```

We just need to make `yaptide/admin/db_manage.py` an executable and add python shebang.

kacper1409:
This is a major rework of simulator upload and download method.
Fluka part done by @kacper1409 , then refactor by @grzanka

Most of the secrets and enviroment needed to configure S3 access for Github Actions were adjusted

This PR introduces:
  - method to download encrypted Fluka binary from S3 
  - splitted `yaptide/admin/simulators.py` into two smaller files as the codebase grew too large
  - created separated commands to download each of the simulator, as keeping code common for all of them meant to much complexity
  - added commands to encrypt and decrypt files locally
  - removed obsolete simulator installation code from celery tasks
  - added pre-commit hook config and dependency


yaptide/admin/simulators.py
```diff
@@ -343,6 +357,52 @@ def download_topas_from_s3(path: Path,
     click.echo(f"Installed Geant4 files into {geant_files_path}")
     return True
 
+def download_fluka_from_s3(path: Path,
```
grzanka: This seems duplicating part of the same code as for SHIELD-HIT12A and TOPAS. Can you extract common code to some separate functions ?

Fluka is also proprietary code, we should make some additional protections in case S3 credentials would leak.
I recommend to encrypt Fluka binaries in the same way as we did for SHIELD-HIT12A

.github/workflows/test-run.yml
```diff
@@ -19,7 +23,7 @@ jobs:
         python-version: ['3.9', '3.10']
         platform: [ubuntu-latest, windows-latest]
         exclude:
-          - python-version: '3.9'
+          - python-version: '3.10'
```
hendzeld: Why?

Does I understand this correctly and this excludes Python 3.10 on Windows from the matrix?
Why was want to exclude python 3.10 instead of 3.9?

README.md
```diff
@@ -93,12 +93,12 @@ curl -i -X DELETE http://localhost:5000/auth/logout
 
 Now registering, updating and deleting users is available with the use of `db_manage.py` located in `yaptide/admin` folder.
 Once docker compose is running, you can use the following command:
-`docker exec -w /usr/local/app/ yaptide_flask python3 yaptide/admin/db_manage.py --help`.
+`docker exec -w /usr/local/app/ yaptide_flask ./yaptide/admin/db_manage.py --help`.
```
hendzeld, line 96: Please verify if this works as intended

I tested it and got:
```log
OCI runtime exec failed: exec failed: unable to start container process: exec: "./yaptide/admin/db_manage.py": permission denied: unknown
```

You cannot execute file if it doesn't have execubatble flag set.
Either by issuing `chmod u+x` and staging file in git repo or by running in in docker.

You can see example  in Dockerfile-worker
```
# Make the scripts executable
# skipcq: DOK-W1001
RUN chmod +x run_worker.sh yaptide/admin/simulators.py
```

Another option is `docker exec -w /usr/local/app/ yaptide_flask python -m yaptide.admin.db_manage --help`
It uses -m (module) to specify pathg to python module that should be executed

README.md
```diff
@@ -93,12 +93,12 @@ curl -i -X DELETE http://localhost:5000/auth/logout
...
 
 To add an user run:
 
 ```bash
-docker exec -w /usr/local/app/ yaptide_flask python3 yaptide/admin/db_manage.py add-user admin --password mysecretpassword
+docker exec -w /usr/local/app/ yaptide_flask ./yaptide/admin/db_manage.py add-user admin --password mysecretpassword
```
hendzeld, line 101: Same as before

tests/test_download_shieldhit.py
```diff
@@ -4,31 +4,31 @@
...
-        result = False
-    if not 'S3_SECRET_KEY' in os.environ:
-        result = False
-    if not 'S3_SHIELDHIT_BUCKET' in os.environ:
-        result = False
-    if not 'S3_SHIELDHIT_KEY' in os.environ:
-        result = False
+    for var_name in ['S3_ENDPOINT', 'S3_ACCESS_KEY', 'S3_SECRET_KEY', 'S3_SHIELDHIT_BUCKET', 'S3_SHIELDHIT_KEY']:
+        if var_name not in os.environ:
+            logging.error(' variable %s not set', var_name)
```
hendzeld: Removing probably not wanted space at the beginning
```suggestion
            logging.error('variable %s not set', var_name)
```

tests/test_download_topas.py
```diff
@@ -4,60 +4,68 @@
...
         if var_name not in os.environ:
             logging.error(' variable %s not set', var_name)
             result = False
     return result
 
+
 @pytest.mark.skipif(sys.platform == "win32", reason="TOPAS does not work on Windows.")
 def test_if_topas_downloaded(tmpdir):
     """Check if TOPAS is downloaded and can be executed"""
     if check_if_environment_variables_set():
```
hendzeld, line 26: Maybe it will be better to reverse this statement?
```suggestion
    if not check_if_environment_variables_set():
      pytest.fail("Required simulator variables were not set properly")
    else:
```

The rest of the code will be the happy case then, so we can drop `else`

Or maybe it will be better to use `pytest.fail` in fixture itself @grzanka?
Small article here: https://medium.com/opsops/pytest-fail-in-a-fixture-and-in-a-test-2122fab0564

yaptide/admin/simulators.py
```diff
@@ -42,437 +26,210 @@ class SimulatorType(IntEnum):
...
+                            help='encryption password')(func)
+        func = click.option('--salt',
+                            type=click.STRING,
+                            envvar='S3_ENCRYPTION_SALT',
+                            default=password,
+                            required=required,
+                            help='encryption salt')(func)
+        return func
+
+    return decorator
```
hendzeld: I like this approach.

Also, this can be rewritten as:

```suggestion
def encryption_options(required: bool = False):
    """Collection of options for encryption"""

    def decorator(func):
        """Decorator for encryption options"""

        @click.option('--password',
                      type=click.STRING,
                      envvar='S3_ENCRYPTION_PASSWORD',
                      default=password,
                      required=required,
                      help='encryption password')
        @click.option('--salt',
                      type=click.STRING,
                      envvar='S3_ENCRYPTION_SALT',
                      default=password,
                      required=required,
                      help='encryption salt')
        @wraps(func)
        def wrapper(*args, **kwargs):
            return func(*args, **kwargs)
        
        return wrapper

    return decorator
```

and with importing `from functools import wraps`


Same applies to `s3credentials` decorator.
Choose one that you like better

More about `@wrap` https://www.geeksforgeeks.org/python-functools-wraps-function/

.github/workflows/test-run.yml
grzanka: You are correct, that was my mistake. The change should be reverted

README.md
grzanka: I think we should add python shebang and executable bit

grzanka: Fixed in 328b90a940c9290f4f926d3b33dfc73e15c5177e

README.md
grzanka: Fixed in 328b90a940c9290f4f926d3b33dfc73e15c5177e

.github/workflows/test-run.yml
grzanka: Fixed in 69226725833437b089ed274e932dfc04d7ef39f0

tests/test_download_topas.py
grzanka: Here my idea was to skip the test if the environment variables are not set, instead of failing it.
This may be the case of local developer without S3 credentials.
The backend can work fine without S3 env variables, in such case demo version of SHIELD-HIT12A (and probably Fluka fake simulator as well) can be used.

yaptide/admin/simulators.py
grzanka: Indeed wrap is more elegant, adapted it in 21a809a8cca4e9c8a69ad4d875e4b8dacbb3d620

grzanka:
@hendzeld I've replied to all the review comments, consider re-reviewing it

hendzeld (approved):

kacper1409:
@grzanka I reviewed Your changes (I cannot approve as assignee). Should I merge this PR?

grzanka (approved):

grzanka:
> @grzanka I reviewed Your changes (I cannot approve as assignee). Should I merge this PR?

I've merged it now

PR #666, opened 2023-10-18
Update openapi.yaml

p1003:


grzanka:
Can you provide good examples of the paramaters for these endpoints ? In our case job id has quite some special form, it would be informative to include it in the documentation.

grzanka:
Does it mean that input_json should be empty ?

image

Write some meaningful text that this example is not complete and that pretty large input json should go there.

Also - why ntasks is equal to zero ?

grzanka:
Explain if this endpoint removes the jobs from the database:

image

We should consider removing the jobs from DB as well. How this could be achieved ? Some addiional parameter to DELETE ?

p1003:
Deletion from database might be added in `DELETE` method from `/jobs` endpoint. In that way it can work for either Celery or SLURM simulations. We can add condition, that it can be done only for simulations which are CANCELED, FAILED or COMPLETED.

@grzanka make issue for that not to leave it here like this.

grzanka:
@p1003 

can you say whats in this endpoint

image

description says nothing about logfiles.

When I make a request to this endpoint, then the logfiles for a task or for a whole simulation will be updated ?

yaptide/static/openapi.yaml
```diff
@@ -410,7 +411,7 @@ paths:
                 $ref: '#/components/schemas/BasicResponse'
   /jobs/direct:
     post:
-      summary: Allows for submitting jobs
+      summary: Allows for submitting jobs to run directly on server.
```
grzanka, line 414: which server ? `ares.cyfronet.pl` ?
be more specific that we mean `dedicated machine in cloud`.
Mention that here the job doesn't wait in the queue as with SLURM.

yaptide/static/openapi.yaml
```diff
@@ -632,8 +634,8 @@ paths:
               schema:
                 $ref: '#/components/schemas/BasicResponse'
     get:
-      summary: Allows for getting job status and additional job's info.
-      description: This endpoint can be used only by keycloak authenticated users
+      summary: Allows for getting job status and additional job's info. Available only for jobs run on SLURM.
```
grzanka, line 637: What is addiional job's info ?

grzanka (approved):

PR #696, opened 2023-11-01
667 commiting not empty dot env files

### LINKED ISSUE #667: Prevent pushing not empty .env file
Previously, one of the contributors pushed `.env` file with credentials.

To prevent this from happening, we could:
- Add a pre-commit hook. It will abort the push or ask the user (if it is possible in hooks) if he really wants to push `.env` file.
- remove and .env file
- other solution

hendzeld:
Resolves #667


.pre_commit_hooks/check_not_empty_files.py
```diff
@@ -0,0 +1,25 @@
+"""Check that files are not empty."""
```
grzanka: Add python shebang

grzanka (approved):

PR #575, opened 2023-09-05
general rework of simulation and worker execution

grzanka:
This PR introduces several improvements to the method which executes simulator binaries:
 - we get rid of eventlet.spawn which simplifies testing and logging
 - common code may be reused by shieldhit and Fluka to spawn simulation process
 - fixes problems with reporting requested and simulated number of primaries, especially for very short jobs

PR #706, opened 2023-11-08
653 progress bar for fluka

### LINKED ISSUE #653: Progress bar for Fluka

hendzeld:
Resolves 653


grzanka:
Note that https://github.com/yaptide/yaptide/pull/575 was merged

grzanka:
Still draft ?

yaptide/celery/utils/progress/fluka_monitor.py
```diff
@@ -0,0 +1,133 @@
...
+                        varbose: bool = False) -> None:
+    """Function reading the fluka output file and raporting progress to the backend."""
+    in_progress = False
+    requested_primaries = 0
+    update_time = 0
+    for line in linte_iterator:
+        if event.is_set():
+            return
+        utc_now = time_now_utc()
+        if re.match(S_OK_OUT_START, line):
```
grzanka:
```suggestion
        if re.match(S_OK_OUT_START, line):
```

I suspect that matching lines based on regexp may be costly. What about using `startstwith` method of a string ?
I think it should be worth to examine what is faster.

Another thing is that the order of regexp in the loop is important. If you have a line which you expect to be in the first part of the file and will appear only once, put it at the end of the loop. First should be the lines that you expect to be most frequent.

yaptide/celery/utils/pymc.py
```diff
@@ -293,6 +294,59 @@ def read_file(event: threading.Event,
     send_task_update(simulation_id, task_id, update_key, up_dict)
 
 
+def read_fluka_file(event: threading.Event,
+                    dirpath: Path,
+                    simulation_id: int,
+                    task_id: str,
+                    update_key: str,
+                    timeout_wait_for_file: int = 20,
+                    timeout_wait_for_line: int = 5 * 60,
```
grzanka: Can you add `_s` suffix to the variable name to denote which time unit we deal with ?

yaptide/celery/utils/pymc.py
```diff
@@ -81,7 +82,7 @@ def command_to_run_fluka(dir_path: Path, task_id: str) -> list[str]:
         input_path=dir_path,  # skipcq: PYL-W0612 # usefull
         simulator_type=SimulatorType.fluka,
         simulator_exec_path=None,  # useless, we guess from PATH
-        cmdline_opts="")  # useless, not useful for fluka
+        cmdline_opts="-M 1")  # useless, not useful for fluka
```
grzanka: why this option ?

hendzeld: I will add an appropriate comment.

About this option:
Fluka by default executes  5 simulations one after another.
From fluka -h (help) sommand
```
  This operation is repeated from the cycle after the one indicated
       on the -N option (default 0) to the one indicated on the -M
       option (default 5).
...

       -N<last_cycle>   number of the previous run cycle number
       -M<final_cycle>  final run number
```

grzanka: Add a comment please. Usually we run multiple cycles when working in command line. That gives us error estimation when running on single CPU

yaptide/celery/tasks.py
```diff
@@ -273,3 +286,29 @@ def monitor_shieldhit(event: threading.Event, tmp_work_dir: str, task_id: int, u
 
     logging.info("No monitoring processes started for task %d", task_id)
     return None
+
+
+def monitor_fluka(event: threading.Event, tmp_work_dir: str, task_id: int, update_key: str,
+                  simulation_id: int) -> Optional[MonitorTask]:
+    """Function running the monitoring process for SHIELDHIT simulation"""
```
grzanka:
```suggestion
    """Function running the monitoring process for Fluka simulation"""
```

yaptide/celery/utils/progress/fluka_monitor.py
```diff
@@ -0,0 +1,129 @@
...
+S_OK_OUT_INIT = "Total time used for initialization:"
+S_OK_OUT_START = "1NUMBER OF BEAM"
+S_OK_OUT_IN_PROGRESS = " NEXT SEEDS:"
+S_OK_OUT_COLLECTED = " All cases handled by Feeder"
+S_OK_OUT_FIN_PATTERN = re.compile(r"^ \* ======(?:( )*?)End of FLUKA [\w\-.]* run (?:( )*?) ====== \*")
+
+logger = logging.getLogger(__name__)
+
+
+def parse_progress_remaining_line(line: str) -> Optional[Tuple[int, int]]:
```
grzanka:
```suggestion
def parse_progress_remaining_line(line: str) -> Optional[tuple[int, int]]:
```

I think we can stick to python > 3.9

grzanka (changes requested):
fix deepsource issues

grzanka (approved):

PR #709, opened 2023-11-10
plgrid-specific access to SHIELD-HIT12A binary

### LINKED ISSUE #686: Problem retrieving results from SLURM when directory is missing
Documentation on https://yaptide.github.io/yaptide/swagger/ doesn't say anything about it:
image

I have one job for which directory with output was removed (or never generated?) from SLURM. This is the error I get on UI:
image

Nothing is presented to the user, but I see POST request with 500 status code.

When I login via ssh to Ares I see indeed status code 1 (failure) for `ls -f ...` command.

```
[ares][plgkongruencj@login01 ~]$ ls -f /net/ascratch/people/plgkongruencj/yaptide_runs/1698676594102920/output | grep .json
[ares][plgkongruencj@login01 ~]$ echo $?
1
[ares][plgkongruencj@login01 ~]$ ls /net/ascratch/people/plgkongruencj/yaptide_runs/1698676594102920/output 
```

The relevant code is here: https://github.com/yaptide/yaptide/blob/113781666fff8bf54bb9eebd6052d603ae620cf4/yaptide/batch/batch_methods.py#L197

also take a look here:
https://github.com/yaptide/yaptide/blob/113781666fff8bf54bb9eebd6052d603ae620cf4/yaptide/routes/common_sim_routes.py#L165

I think we should have some level of error checking on Fabric command and a way to communicate that to the User.

grzanka:
Seems to work as right now SHIELD-HIT12A is available in the module system on Ares.
I tested it on yap-dev and was able to receive simulation output from Ares

### DIFF
--- a/yaptide/batch/string_templates.py
+++ b/yaptide/batch/string_templates.py
@@ -1,34 +1,12 @@
 SUBMIT_SHIELDHIT: str = """#!/bin/bash
 OUT=`mktemp`
-export PATH="$PATH:$PLG_GROUPS_STORAGE/plggccbmc"
-module load gcc/11.3.0
+module load shieldhit
 
 ROOT_DIR={root_dir}
-BIN_DIR=$ROOT_DIR/bin
 cd $ROOT_DIR
 mkdir -p $ROOT_DIR/workspaces/task_{{0001..{n_tasks}}}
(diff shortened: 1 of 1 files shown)


hendzeld (approved):
LGTM

PR #804, opened 2023-12-29
Update ghcr_packages.md

grzanka:
Codespaces doesn't require any login to GHCR.io, what about gitpod ?


macsko:
In my case gitpod requires to log in

grzanka:
> In my case gitpod requires to log in

True, then you need to be more specific in documentation, explain where login is needed and when not

macsko:
@grzanka I created a new PR applying your comment #810

PR #816, opened 2024-01-05
Update converter module

hendzeld:


hendzeld:
Probably we should also add new version of pymchelper after its release.

grzanka (approved):

PR #825, opened 2024-01-11
yaptide openshift integration

ChrisGadek1:


nginx.conf
```diff
@@ -1,6 +1,6 @@
 # Flask backend is listening on port 6000 on host called yaptide_flask
 upstream backend_flask {
-    server yaptide_flask:6000;
+    server yaptide-4.yaptide.svc.cluster.local:6000;
```
grzanka: This one should be done in much smarter way, as it will probably break the way app is deployed using docker compose.

Dockerfile-flask
```diff
@@ -5,7 +5,10 @@ EXPOSE 6000
 
 # set environment variables
 ENV FLASK_APP=yaptide.application
-WORKDIR /usr/local/app
+WORKDIR /home
+
+RUN python -m venv /home/venv && chmod -R 777 /home/venv
+ENV PATH="/home/venv/bin:$PATH"
```
grzanka: Is this really the only way to install pip packages in venv ?
have you tried any other ways, like `pip install --user` ?

ChrisGadek1: `pip install --user` seems to not work, because open shift containers don't instantiate home directory for user, so `USER_SITE` does not exist. Also, it tries to create packages for root user, because all command in Openshift in dockerfile level are run as the root user, so it doesn't make sense.  I will try to change the owner of the venv directory to the container user and I think it will be the best I can do.

grzanka: OK, thanks for explanation !

grzanka (changes requested):
Can you revert changes of the converter submodule ?

ChrisGadek1:
> Can you revert changes of the converter submodule ?

How can I do it? I didn't find how to remove submodules from PR

grzanka:
Postponing till more quiet times

PR #836, opened 2024-01-21
Tasks progress

michalskorek:


grzanka:
@jagodek more conflicts coming in, I've just merged some of the bumps

jagodek:
> @jagodek more conflicts coming in, I've just merged some of the bumps

ok. I pushed my commits by missclick, so i'm not done yet

grzanka:
> > @jagodek more conflicts coming in, I've just merged some of the bumps
> 
> ok. I pushed my commits by missclick, so i'm not done yet

Commit and push as often as you can, not waiting till being ready :) I really like PRs with >100 of commits.

grzanka:
@jagodek take a look here: https://github.com/yaptide/deploy/blob/120-describe-interaction-between-frontend-and-backend-when-user-submits-simulation/docs/sequence.md

that may help you to get what was the state of update mechanism before this PR

grzanka:
@jagodek as fot testing this issue, maybe this solution may solve some problems:
https://github.com/yaptide/yaptide/issues/973
?

jagodek:
@grzanka I've delved more into the issue. The main problem is that scheduler doesn't work in pytest. I also noticed that FLASK_DEBUG=True in pytest.ini should be changed to False or disabled in another way ( it's something about flask_apscheduler library and debug mode ). But this little progress doesn't solve issue.

jagodek:
also can't tell if job is running during test_run_simulation because logs are only displayed after test is done

jagodek:
Basically  persistence/db_methods don't work in scheduled task

grzanka:
> Basically persistence/db_methods don't work in scheduled task

Can you be more specific on `don't work` ?
What is the issue ? If you start the app in developer mode (I am not speaking about tests) and submit simulation with 100 parallel tasks, does scheduler communicate with redis ? Does it communicate with API ?

jagodek:
Everything about communication with redis is fine.

jagodek:
It's as if db_methods called in scheduled task during test act as if db was empty i guess.

PR #872, opened 2024-02-29
Bump eventlet to fix issues with connection to redis

### LINKED ISSUE #871: yaptide_worker cannot connect to redis
Gives:
```
consumer: Cannot connect to redis://redis:6379/0: Error -3 connecting to redis:6379. Lookup timed out..
Trying again in 2.00 seconds... (1/100)
```

grzanka:
I suspected the issues with connection to redis were caused by the way celery is using its redis connector. I've connected to the `yaptide_worker` image and played with simple celery apps which experienced no issues in connecting to redis.
Then I ran simply `celery --app yaptide.celery.worker worker` and discovered no issue.
Then subsequently I was adding more switches. I've realized that the problem appeared on adding `--pool eventlet`. I've inspected on what version `eventlet` library is. Instinctively I've tried updating to the newest one 0.35.2 and this seemed to fix the problem.

Strangle I don't see any development in 0.35.1 and 0.35.2 that seems to be related to our issues:

https://github.com/eventlet/eventlet/blob/v0.35.2/NEWS#L4-L18

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,23 @@
 # yaptide backend
 
-## Installation
+## Getting the code
 
-Run: `$ pip install -r requirements.txt`
+Clone the repository including submodules:
 
-Download/update sumbmodules:
+```shell
+git clone --recurse-submodules https://github.com/yaptide/yaptide.git
+```
+
+In case you have used regular `git clone` command, without `--recurse-submodules` option, you can still download the submodules by running:
 
 ```shell
 git submodule update --init --recursive
 ```
+
+## Installation
+
+Run: `pip install -r requirements.txt`
+
 
 ## Running the app
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ version: "3.4"
 services:
 
     redis:
-        image: redis:6-alpine
+        image: redis:7-alpine
         container_name: yaptide_redis
         healthcheck:
             test: ["CMD", "redis-cli", "ping"]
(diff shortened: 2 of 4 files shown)


grzanka:
OK, most probably its this one:
https://github.com/eventlet/eventlet/issues/913

> As part of addressing a potential resolution DoS, we had to add a new parameter to dns.query.udp(), "ignore_errors". Dnspython's stub resolver in 2.6.0 will pass True for this flag, so the monkeypatched version of dns.query.udp() in eventlet will need to be updated or resolutions will start failing when people install dnspython 2.6.0. I just released 2.6.0rc1, and will release the final version a week from now, so you can test with that. If you just fix the API, then you will be vulnerable to the potential DoS still, as the fix involves ignoring various errors in the receive loop instead of giving up. You can see the revised logic in dnspython's dns.query.receive_udp() method. Pinning eventlet to <2.6.0 would also be a mitigation, though again with the potential DoS.

`dnspython` is not on the list of pinned requirements as its one of child-dependencies.

A month ago it got update to 2.6.0 and then later to 2.6.1:
https://pypi.org/project/dnspython/#history

hendzeld (approved):
Overall, it is fine.

For reference, changed hostnames could be replaced with links: https://docs.docker.com/compose/networking/#link-containers
In our case, it is not a problem.

I once had an issue with some Golang service in docker. Underscores aren't considered a valid character in the hostname part of the URIs. :)

obraz
From https://en.wikipedia.org/wiki/Hostname

grzanka:
> Overall, it is fine.
> 
> For reference, changed hostnames could be replaced with links: https://docs.docker.com/compose/networking/#link-containers In our case, it is not a problem.
> 
> I once had an issue with some Golang service in docker. Underscores aren't considered a valid character in the hostname part of the URIs. :)
> 
> obraz From https://en.wikipedia.org/wiki/Hostname

Good point for a further issue.

PR #1168, opened 2024-10-26
Fix code scanning alert no. 4: Reflected server-side cross-site scripting

grzanka:
Fixes [https://github.com/yaptide/yaptide/security/code-scanning/4](https://github.com/yaptide/yaptide/security/code-scanning/4)

To fix the problem, we need to ensure that any user-provided input included in the response is properly escaped to prevent XSS attacks. In this case, we can use the `html.escape` function from Python's standard library to escape the `message` parameter before including it in the `response_dict`.

- Modify the `yaptide_response` function in `yaptide/routes/utils/response_templates.py` to escape the `message` parameter.
- Import the `html` module to use the `html.escape` function.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/yaptide/routes/utils/response_templates.py
+++ b/yaptide/routes/utils/response_templates.py
@@ -1,5 +1,7 @@
 from flask import Response, make_response
 import html
+
+
 def yaptide_response(message: str, code: int, content: dict = None) -> Response:
     """Function returning Response object"""
     response_dict = {'message': html.escape(message)}
--- a/yaptide/routes/utils/response_templates.py
+++ b/yaptide/routes/utils/response_templates.py
@@ -1,9 +1,9 @@
 from flask import Response, make_response
-
+import html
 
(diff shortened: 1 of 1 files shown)


yaptide/routes/utils/response_templates.py
```diff
@@ -1,9 +1,9 @@
 from flask import Response, make_response
-
+import html
```
jagodek, line 3:
```suggestion


```

jagodek: deepsource

jagodek (dismissed):
deepsource. works good

grzanka:
> deepsource. works good

I commited your fix. Can you review again?

jagodek (dismissed):

jagodek:
now there are no blank lines before yaptide_response at all

jagodek (approved):

PR #1196, opened 2024-11-12
Fixing cancelling simulation for celery

### LINKED ISSUE #1173: Error on cancelling celery simulation
When number of tasks is greater than number of cores for celery_worker there is error after hitting cancel button. Sometimes cancelling works when number of tasks is greater just by few, but if it's more then it won't work.

jagodek:
The idea is to look into redis queue and remove tasks related to running simulation before function cancelling executing tasks happen

### DIFF
--- a/yaptide/celery/utils/manage_tasks.py
+++ b/yaptide/celery/utils/manage_tasks.py
@@ -13,6 +13,7 @@ from yaptide.utils.enums import EntityState
 
 
 def stop_tasks_in_worker(simulation):
+    """Stops tasks that are running or are reserved in celery"""
     ids_in_worker = get_tasks_from_celery(simulation_id=simulation.id)
     task_ids = [task['task_id'] for task in ids_in_worker]
     celery_ids = [task['celery_id'] for task in ids_in_worker]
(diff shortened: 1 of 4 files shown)


jagodek:
For now there is a problem with getting from redis task object. Inside task object the nested dictionary kwargsrepr contains string that is some shortened version of args passed to run_single_simulation which ends with  ...', ...}}. Somewhere behind these dots is simulation_id but I can't find a way to get full kwasrgsrepr dictionary.

jagodek:
Managed to decode simulation_id from task object in redis so deleting tasks from redis works. Because some decoding is done i'll test performance. What's concerning that I noticed that not all task that have started executing have their celery_id saved in database.

yaptide/celery/utils/manage_tasks.py
```diff
@@ -1,13 +1,68 @@
...
+        item_data = json.loads(item)
+        body_encoded = item_data["body"]
+        decoded_body = base64.b64decode(body_encoded)
+        body_data = json.loads(decoded_body)
+        kwargs = body_data[1]
+        array_ids.append({"task_id": int(kwargs.get("task_id")), "simulation_id": int(kwargs.get("simulation_id"))})
+    return array_ids
+
+
+def delete_tasks_from_redis(simulation_id):
```
grzanka, line 46: what if we decide to reconfigure redis and use another (other than redis) storage backend ?

PR #1224, opened 2024-11-22
CI rework to handle S3 connectivity issues

grzanka:
Extracted simulator downloads to a single action with a fallback to demo version download if S3 fails


jagodek (approved):
So there still is some problem with s3 and for now there's fallback to demo version of shieldhit?

grzanka:
> So there still is some problem with s3 and for now there's fallback to demo version of shieldhit?

Yes and it seems the issue is on GitHub side

PR #1240, opened 2024-12-05
improved SH12A demo exe download

### LINKED ISSUE #1239: Get SH12A demo test on GitHub action fails
Screenshot 2024-12-05 at 14 19 04

grzanka:
Previously, detection of the step failure with `if [[ ${{ job.status }} == 'failure' ]]; then` was broken, this is a version which seems to work. Even if download via S3 failed, the "demo download" step was never activated. In case of S3 download failure, the `job.status` was always "success". Below AI generated summary of the PR:

image

This pull request includes updates to the `.github/workflows/test-run.yml` file to improve the logic for handling simulator downloads on Ubuntu.

Changes to workflow logic:

* Added an `id` to the "Get simulators for Linux" step to track its outcome for conditional execution.
* Modified the condition for the "Get demo simulators for Linux due to timeout" step to check the outcome of the previous step instead of always running on failure.

### DIFF
--- a/.github/workflows/test-run.yml
+++ b/.github/workflows/test-run.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Get simulators for Linux
         if: matrix.platform == 'ubuntu-latest'
         run: |
-            poetry run python yaptide/admin/simulators.py download-shieldhit --decrypt --dir bin/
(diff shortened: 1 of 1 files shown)


jagodek (approved):
lgtm

PR #1243, opened 2024-12-07
Adjust estimator db model

SzymanskiBartlomiej:
This pull request includes several important changes to the database schema, migration scripts, and various parts of the codebase to handle a new `file_name` attribute for `EstimatorModel`. The changes also include updates to tests, routes, and utility functions to accommodate the new attribute.

### Database Schema and Migration:
* Added a new `file_name` column to the `EstimatorModel` table to store the name extracted from files generated by the simulator.
* Created a migration script to add the `file_name` column, populate it, and update existing records accordingly.

### Codebase Updates:
* Updated the `fetch_estimator_by_sim_id_and_file_name` function to fetch estimators using the new `file_name` attribute.
* Modified the `send_simulation_results` function to sort files before processing them.

### Tests:
* Updated tests to include the new `file_name` attribute when creating and validating `EstimatorModel` instances. [[1]](diffhunk://#diff-82fbc96aa3f082d4667c09a2a35aec050c73120de9174e1f37bef26ef9cd3115L361-R370) [[2]](diffhunk://#diff-82fbc96aa3f082d4667c09a2a35aec050c73120de9174e1f37bef26ef9cd3115L383-R386)

### Routes and Utilities:
* Updated route handlers and utility functions to use the new `file_name` attribute for fetching and updating estimators. [[1]](diffhunk://#diff-40ee953c21b980818ac611a5e33f1dbe7220a3d9bae1341b2556feb8fa1ee157L10-R20) [[2]](diffhunk://#diff-40ee953c21b980818ac611a5e33f1dbe7220a3d9bae1341b2556feb8fa1ee157R111) [[3]](diffhunk://#diff-40ee953c21b980818ac611a5e33f1dbe7220a3d9bae1341b2556feb8fa1ee157R142-R167) [[4]](diffhunk://#diff-40ee953c21b980818ac611a5e33f1dbe7220a3d9bae1341b2556feb8fa1ee157R197-R221)

### Dependency Updates:
* Updated the `pymchelper` dependency version in `pyproject.toml`.

PR #1171, opened 2024-10-28
1137 adds new endpoints to allow quering results for single page

SzymanskiBartlomiej:
This pull request introduces several changes to the database schema, API endpoints, and test cases to support additional metadata for pages and enhance the querying capabilities. The most important changes include the addition of new columns to the `Page` model, updates to the API endpoints to handle page-specific queries, and modifications to the test cases to validate these new functionalities.

### Database Schema Updates:
* [`migrations/versions/5003b9acb1f4_.py`](diffhunk://#diff-764071cc37ed8a4e6ae1b267c6c7114bebc331e652859b3f648c5b861a1225c2R1-R53): Added `page_name` and `page_dimension` columns to the `Page` table and included a migration script to populate these columns based on existing data.
* [`yaptide/persistence/models.py`](diffhunk://#diff-4b8473e45fecb31924a9768d0ea11ecd57ea27d99f39689517035d22a9c2689eR362-R366): Updated the `PageModel` class to include `page_name` and `page_dimension` attributes.

### API Enhancements:
* [`yaptide/routes/common_sim_routes.py`](diffhunk://#diff-40ee953c21b980818ac611a5e33f1dbe7220a3d9bae1341b2556feb8fa1ee157L160-R183): Enhanced the `ResultsResource` class to support querying results by specific page numbers or ranges. Added a helper function `parse_page_numbers` to parse page ranges from query parameters. [[1]](diffhunk://#diff-40ee953c21b980818ac611a5e33f1dbe7220a3d9bae1341b2556feb8fa1ee157L160-R183) [[2]](diffhunk://#diff-40ee953c21b980818ac611a5e33f1dbe7220a3d9bae1341b2556feb8fa1ee157R251-R252) [[3]](diffhunk://#diff-40ee953c21b980818ac611a5e33f1dbe7220a3d9bae1341b2556feb8fa1ee157R261) [[4]](diffhunk://#diff-40ee953c21b980818ac611a5e33f1dbe7220a3d9bae1341b2556feb8fa1ee157R271-R272) [[5]](diffhunk://#diff-40ee953c21b980818ac611a5e33f1dbe7220a3d9bae1341b2556feb8fa1ee157L262-R300)
[...]

PR #1244, opened 2024-12-08
251 implement method to cancel running jobs and fetching results of unfinished tasks

### LINKED ISSUE #251: implement method to cancel running jobs (and fetching results of unfinished tasks)

NiEkOKsU:


dump.rdb
grzanka, line 1: what is that file ?

grzanka:
https://github.com/yaptide/yaptide/pull/1243 was merged now, good idea to rebase or merge

dump.rdb
NiEkOKsU: It is some trash file created by redis, I will remove it in next commit. Maybe it is a good idea to add dump.rdb to a .gitignore?

grzanka: please remove it and add to .gitignore.

yaptide/persistence/models.py
```diff
@@ -196,6 +196,10 @@ class TaskModel(db.Model):
                                         nullable=False,
                                         default=EntityState.PENDING.value,
                                         doc="Task state (i.e. 'pending', 'running', 'completed', 'failed')")
+    sim_pid: Column[int] = db.Column(db.Integer,
```
grzanka: you need a flask migrate stuff for that !

yaptide/routes/celery_routes.py
```diff
@@ -165,14 +166,20 @@ def delete(user: UserModel):
...
         # We don't want it to run accidentally.
-        celery_app.control.revoke(simulation.merge_id, terminate=True, signal="SIGINT")
-        celery_app.control.revoke(celery_ids, terminate=True, signal="SIGINT")
-        update_simulation_state(simulation=simulation, update_dict={"job_state": EntityState.CANCELED.value})
-        for task in tasks:
-            if task.task_state in [EntityState.PENDING.value, EntityState.RUNNING.value]:
-                update_task_state(task=task, update_dict={"task_state": EntityState.CANCELED.value})
-
-        terminate_unfinished_tasks.delay(simulation_id=simulation.id)
+        if simulation.sim_type == "shieldhit" and simulation.job_state in (EntityState.RUNNING.value):
```
grzanka: check this enum: https://github.com/yaptide/yaptide/blob/77a67dad5c9e74d3e4b76ef9adbabc354ed3b9ef/yaptide/utils/enums.py#L72

yaptide/routes/celery_routes.py
```diff
@@ -165,14 +166,20 @@ def delete(user: UserModel):
...
+            logging.debug('Inside cancelling shieldhit')
+            for task in tasks:
+                logging.debug("PID: %d", task.sim_pid)
+                subprocess.run(['kill', '-s', 'SIGINT', str(task.sim_pid)], check=True)
```
grzanka:
```suggestion
                subprocess.run(['kill', '--signal', 'SIGINT', str(task.sim_pid)], check=True)
```

yaptide/routes/celery_routes.py
```diff
@@ -165,14 +166,20 @@ def delete(user: UserModel):
...
-        celery_app.control.revoke(celery_ids, terminate=True, signal="SIGINT")
-        update_simulation_state(simulation=simulation, update_dict={"job_state": EntityState.CANCELED.value})
-        for task in tasks:
-            if task.task_state in [EntityState.PENDING.value, EntityState.RUNNING.value]:
-                update_task_state(task=task, update_dict={"task_state": EntityState.CANCELED.value})
-
-        terminate_unfinished_tasks.delay(simulation_id=simulation.id)
+        if simulation.sim_type == "shieldhit" and simulation.job_state in (EntityState.RUNNING.value):
+            logging.debug('Inside cancelling shieldhit')
+            for task in tasks:
```
grzanka: `man kill` says you can kill multiple pid's at once:

image

grzanka: `man kill` says you can kill multiple pid's at once:

image

grzanka:
@NiEkOKsU fix merge conflicts

migrations/versions/bb893eb0d961_.py
```diff
@@ -0,0 +1,52 @@
...
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.orm import sessionmaker
+
+from yaptide.persistence.models import TaskModel
+
+
+# revision identifiers, used by Alembic.
+revision = 'bb893eb0d961'
+down_revision = '41149fa64bd2'
```
grzanka: Take a look at recently merged PR:
https://github.com/yaptide/yaptide/pull/1171/files#diff-764071cc37ed8a4e6ae1b267c6c7114bebc331e652859b3f648c5b861a1225c2R15

the last revision is now `5003b9acb1f4` I think.

grzanka: Re-create this script

migrations/versions/d1a281f6f2dd_.py
```diff
@@ -0,0 +1,51 @@
...
+revision = 'd1a281f6f2dd'
+down_revision = '5003b9acb1f4'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('Task', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('sim_pid', sa.Integer(), nullable=False, server_default=sa.text("-1")))
```
grzanka: Why don't you assume these columns to be nullable ?

yaptide/batch/batch_methods.py
```diff
@@ -309,19 +309,21 @@ def get_job_results(simulation: BatchSimulationModel, user: KeycloakUserModel, c
...
     collect_id = simulation.collect_id
 
     try:
         con = get_connection(user=user, cluster=cluster)
-
-        con.run(f'scancel {array_id}')
-        con.run(f'scancel {collect_id}')
-        con.run(f'rm -rf {job_dir}')
+        if to_fetch:
+            con.run(f'scancel --signal SIGINT {array_id}')
```
grzanka: What about killing merge job if we fetch results?

yaptide/routes/celery_routes.py
```diff
@@ -165,14 +168,36 @@ def delete(user: UserModel):
...
+            file_name = '/rfluka.stop'
+            for task in tasks:
+                entries = os.listdir(task.path_to_sim)
+                target_folder = [
+                    entry for entry in entries
+                    if entry.startswith('fluka_') and os.path.isdir(os.path.join(task.path_to_sim, entry))
+                ][0]
+                target_path = os.path.join(task.path_to_sim, target_folder)
+                with open(target_path + file_name, 'w') as file:
+                    file.write("")
```
grzanka: maybe these lines could be extracted to a separate function, like `stop_fluka_simulation` with necessary parameteres ?

yaptide/routes/celery_routes.py
```diff
@@ -165,14 +168,36 @@ def delete(user: UserModel):
...
+                EntityState.RUNNING.value):
+            logging.debug('Cancelling fluka with fetching data')
+            file_name = '/rfluka.stop'
+            for task in tasks:
+                entries = os.listdir(task.path_to_sim)
+                target_folder = [
+                    entry for entry in entries
+                    if entry.startswith('fluka_') and os.path.isdir(os.path.join(task.path_to_sim, entry))
+                ][0]
+                target_path = os.path.join(task.path_to_sim, target_folder)
```
grzanka: use `pathlib` instead of `os.path.join`

yaptide/routes/celery_routes.py
```diff
@@ -165,14 +168,36 @@ def delete(user: UserModel):
...
+                command_as_list.append(str(task.sim_pid))
+            logging.debug(command_as_list)
+            subprocess.run(command_as_list, check=True)
+        elif to_fetch is True and simulation.sim_type == SimulationType.FLUKA.value and simulation.job_state in (
+                EntityState.RUNNING.value):
+            logging.debug('Cancelling fluka with fetching data')
+            file_name = '/rfluka.stop'
+            for task in tasks:
+                entries = os.listdir(task.path_to_sim)
+                target_folder = [
```
grzanka: maybe using `glob` from `pathlib` would simplify the code ?

yaptide/routes/celery_routes.py
```diff
@@ -165,14 +168,36 @@ def delete(user: UserModel):
...
+            command_as_list = ['kill', '--signal', 'SIGINT']
+            for task in tasks:
+                command_as_list.append(str(task.sim_pid))
+            logging.debug(command_as_list)
+            subprocess.run(command_as_list, check=True)
+        elif to_fetch is True and simulation.sim_type == SimulationType.FLUKA.value and simulation.job_state in (
+                EntityState.RUNNING.value):
+            logging.debug('Cancelling fluka with fetching data')
+            file_name = '/rfluka.stop'
+            for task in tasks:
```
grzanka: why don't you check for a state of each individual task ? some of them may be still pending, even in the simulation is running, some of them completed

migrations/versions/d1a281f6f2dd_.py
NiEkOKsU: This columns can be nullable, in addition i will move them to CeleryTaskModel because they are used only by celery simulation

yaptide/batch/batch_methods.py
NiEkOKsU: Why should we kill merge job? As I know if we kill merge job, collect script will not run and data will not be send to the database.

grzanka: We are in `delete_job` function, I would expect it terminates the job completely.
Maybe for better clarity you should define a separate function which would be named `fetch_results_without_stopping` or something that bears some meaning ?

yaptide/routes/celery_routes.py
```diff
@@ -158,21 +161,49 @@ def delete(user: UserModel):
...
+            logging.debug(command_as_list)
+            subprocess.run(command_as_list, check=True)
+            celery_app.control.revoke(celery_ids, terminate=True, signal="SIGINT")
+        elif to_fetch is True and simulation.sim_type == SimulationType.FLUKA.value and simulation.job_state in (
+                EntityState.RUNNING.value):
+            logging.debug('Cancelling fluka with fetching data')
+            file_name = 'rfluka.stop'
+            for task in tasks:
+                if task.task_state == EntityState.RUNNING.value:
+                    target_folder = next(Path(task.path_to_sim).glob("fluka_*/"))
```
grzanka: what if no folder with "fluka_*" pattern exists ?

yaptide/routes/celery_routes.py
```diff
@@ -158,21 +161,49 @@ def delete(user: UserModel):
...
+                    target_path = target_folder / file_name
+                    logging.debug("Path: %s", target_path)
+                    with target_path.open('w') as file:
+                        file.write("")
```
grzanka: Check `Path.write_text` in pathlib

yaptide/routes/batch_routes.py
```diff
@@ -164,17 +165,22 @@ def delete(user: KeycloakUserModel):
                                     })
 
         cluster = fetch_cluster_by_id(cluster_id=simulation.cluster_id)
-
-        result, status_code = delete_job(simulation=simulation, user=user, cluster=cluster)
+        if to_fetch:
+            result, status_code = fetch_results_without_stopping(simulation=simulation,
```
grzanka: why don't you move that functionality to a different endpoint than delete ?
if you fetch results and stop simulation thats OK, but if you fetch results and continue it doesn't make sense to keep that in `delete` enpoint

grzanka: If you plan only to stop simulation and fetch the results then you should name the function not  `fetch_results_without_stopping`, but something like `stop_and_fetch_results`

yaptide/routes/batch_routes.py
```diff
@@ -148,6 +148,7 @@ def delete(user: KeycloakUserModel):
         params_dict: dict = schema.load(request.args)
 
         job_id: str = params_dict["job_id"]
+        to_fetch = params_dict.get('to_fetch', True)
```
grzanka: Maybe `fetch_results` would be a better name than `to_fetch` ?

yaptide/routes/batch_routes.py
NiEkOKsU: So you want to make 2 new endpoints for cancel simulation with data fetching, one for celery and one for batch? I'm asking because making new endpoint only for a batch simulation in my opinion could provide more dificulties than positives.

grzanka: I am just worried about consistency of naming of all the items we have: endpoints, functions etc...
If the only aim of this PR is to introduce the feature of "get the result" on cancel, then why do you name the function `fetch_results_without_stopping` ? 
Don't follow blindly any of my advice, just ensure that the names of variables, functions etc are meaningful and consistent.

grzanka:
Also - there are failing tests and deepsource complaning

yaptide/routes/batch_routes.py
grzanka: If you consider only "cancel and get the results" then one endpoint (delete) should be OK

yaptide/batch/batch_methods.py
```diff
@@ -327,3 +327,27 @@ def delete_job(simulation: BatchSimulationModel, user: KeycloakUserModel,
         return {"message": "Job cancelation failed"}, 500
 
     return {"message": "Job canceled"}, 200
+
+
+def fetch_results_without_stopping(simulation: BatchSimulationModel, user: KeycloakUserModel,
+                                   cluster: ClusterModel) -> tuple[dict, int]:  # skipcq: PYL-W0613
+    """Function that stops simulation without cancelling collection job"""
+    array_id = simulation.array_id
+    tasks = fetch_celery_tasks_by_sim_id(sim_id=simulation.id)
```
grzanka: why do you fetch celery tasks in batch methods ? If I remember correctly batch methods were supposed to run on Ares

yaptide/batch/batch_methods.py
```diff
@@ -327,3 +327,27 @@ def delete_job(simulation: BatchSimulationModel, user: KeycloakUserModel,
...
+    try:
+        con = get_connection(user=user, cluster=cluster)
+        i = 1
+        fetch_str = 'scancel --signal=SIGINT'
+        delete_str = 'scancel'
+        for task in tasks:
+            if task.task_state == EntityState.RUNNING.value:
+                fetch_str += f' {array_id}_{i}'
+            if task.task_state in (EntityState.PENDING.value, EntityState.UNKNOWN.value):
+                delete_str += f' {array_id}_{i}'
```
grzanka, line 348: add some comments what happens here, explaining that you will with SIGINT only running tasks, asking them to dump the results and kill in normal way all pending tasks

yaptide/batch/batch_methods.py
```diff
@@ -327,3 +327,27 @@ def delete_job(simulation: BatchSimulationModel, user: KeycloakUserModel,
         return {"message": "Job cancelation failed"}, 500
 
     return {"message": "Job canceled"}, 200
+
+
+def fetch_results_without_stopping(simulation: BatchSimulationModel, user: KeycloakUserModel,
+                                   cluster: ClusterModel) -> tuple[dict, int]:  # skipcq: PYL-W0613
+    """Function that stops simulation without cancelling collection job"""
```
grzanka, line 334: Maybe in the name you should say something like `cancel_simulation_with_result_dump` saying in docstring that you only cancel simulation (array job) without cancelling merge job which should process all dumped files

yaptide/batch/batch_methods.py
```diff
@@ -327,3 +327,27 @@ def delete_job(simulation: BatchSimulationModel, user: KeycloakUserModel,
...
+        i = 1
+        fetch_str = 'scancel --signal=SIGINT'
+        delete_str = 'scancel'
+        for task in tasks:
+            if task.task_state == EntityState.RUNNING.value:
+                fetch_str += f' {array_id}_{i}'
+            if task.task_state in (EntityState.PENDING.value, EntityState.UNKNOWN.value):
+                delete_str += f' {array_id}_{i}'
+        con.run(fetch_str)
+        con.run(delete_str)
```
grzanka: how long does it take to execute ? if longer than ~0.1s then maybe we should do this stuff in celery helper instance ? just like job submission

yaptide/batch/batch_methods.py
NiEkOKsU: yes, there is used bad function, i will change it to this one dedicated to batch simulation

yaptide/routes/celery_routes.py
```diff
@@ -158,21 +162,16 @@ def delete(user: UserModel):
...
-        # We don't want it to run accidentally.
-        celery_app.control.revoke(simulation.merge_id, terminate=True, signal="SIGINT")
-        celery_app.control.revoke(celery_ids, terminate=True, signal="SIGINT")
-        update_simulation_state(simulation=simulation, update_dict={"job_state": EntityState.CANCELED.value})
-        for task in tasks:
-            if task.task_state in [EntityState.PENDING.value, EntityState.RUNNING.value]:
-                update_task_state(task=task, update_dict={"task_state": EntityState.CANCELED.value})
-
-        terminate_unfinished_tasks.delay(simulation_id=simulation.id)
+        celery_ids = []
```
grzanka: what are celery_ids ?

yaptide/celery/tasks.py
```diff
@@ -246,7 +258,8 @@ def merge_results(results: list[dict]) -> dict:
             "update_key": update_key
         }
         post_update(dict_to_send)
-    for i, result in enumerate(results):
+    total_particles = 0
```
grzanka: better would be `total_primaries`

yaptide/celery/tasks.py
```diff
@@ -257,10 +270,16 @@ def merge_results(results: list[dict]) -> dict:
...
         if averaged_estimators is None:
             averaged_estimators: list[dict] = result.get("estimators", [])
+            total_particles += int(averaged_estimators[0]["metadata"]["number_of_primaries"])
+            logging.debug(total_particles)
             # There is nothing to average yet
             continue
 
-        averaged_estimators = average_estimators(averaged_estimators, result.get("estimators", []), i)
+        new_particles = int(result.get("estimators", [])[0]["metadata"]["number_of_primaries"])
+        logging.debug("Avereging results from %d particles", new_particles)
```
grzanka:
```suggestion
        logging.debug("Averaging results from %d primaries", new_particles)
```

yaptide/celery/tasks.py
```diff
@@ -257,10 +270,16 @@ def merge_results(results: list[dict]) -> dict:
...
+        averaged_estimators = average_estimators(averaged_estimators, result.get("estimators", []), total_particles,
```
grzanka: does it calculate normal average or weighted average ?
in case of interuppted simulation a weighted average would be needed.
Maybe you can (re)use this code: https://github.com/DataMedSci/pymchelper/blob/master/pymchelper/averaging.py

NiEkOKsU: It calculates weighted average

yaptide/routes/celery_routes.py
NiEkOKsU: List of celery_id from CeleryTaskModel, residue from the old version of the code

grzanka:
See https://github.com/yaptide/ui/issues/1905 as well

NiEkOKsU:
Please look at this failing deepsources. Is it necesary to do something with this?

grzanka:
> Please look at this failing deepsources. Is it necesary to do something with this?

Yes, deepsource is quite correct. 

It's good idea to make ARRAY_SHIELDHIT_BASH an r-string (assuming everything would later work, as it will be later treated as f-string).
image

yaptide/celery/utils/progress/fluka_monitor.py
```diff
@@ -153,3 +155,28 @@ def read_fluka_out_file(event: threading.Event,
...
+
+
+def read_fluka_file_offline(filepath: Path) -> tuple[int, int]:
+    """Reads fluka out file and returns number of simulated and requested primaries"""
+    simulated_primaries = 0
+    requested_primaries = 0
+    filepath = filepath / "fl_sim001.out"
+    in_progress = False
+    try:
+        with open(filepath, 'r') as f:
```
grzanka: bad idea to modify function argument by overwriting `filepath`
I think better name for function argument would be `output_dir` or something like that

also the name of the function should suggest what you actually read.
Is it the binary (if I remember correctly "unformatted" in Fluka language) estimator output, i.e. `*_fort*` files ?
Or is it general standard output file ? Take a look here 
http://www.fluka.org/fluka.php?id=man_onl&sub=95 or somewhere else in Fluka documentation and use such naming that user would know what that function reads.

grzanka: As for deepsource, maybe it will stop complain if you pass `output_dir / "fl_sim001.out"` to open ?

yaptide/celery/tasks.py
grzanka: But not all quantities should be averaged, see details on pymchelper code.
An example pymchelper code usage can be found here:
https://github.com/yaptide/yaptide/blob/7a66b1c1d846987f380d6b5ba5d0ab6d7b9e31a0/yaptide/celery/utils/pymc.py#L111

As for weighted average - do you mean this code is doing weighted average ?
https://github.com/yaptide/yaptide/blob/7a66b1c1d846987f380d6b5ba5d0ab6d7b9e31a0/yaptide/celery/utils/pymc.py#L172-L174

If thats the case, then its completely unreadable. It would need a good paragraph of docstring explaining that. 

Ideally accompanied with doctest: https://docs.python.org/3/library/doctest.html

NiEkOKsU: [I have changed this function in this commit. you have linked here a averrage_values from a master.](https://github.com/yaptide/yaptide/blob/251-implement-method-to-cancel-running-jobs-and-fetching-results-of-unfinished-tasks/yaptide/celery/utils/pymc.py#L178-L183)

grzanka: As for now it's ok, but the code is completely unreadable. It's hard to guess from reading the code and comments what it is doing.

yaptide/batch/batch_methods.py
NiEkOKsU: I doubt it, but it needs to be checked on yap-dev

yaptide/celery/utils/manage_tasks.py
```diff
@@ -87,3 +92,54 @@ def translate_celery_state_naming(job_state: str) -> str:
         return EntityState.COMPLETED.value
     # Others are the same
     return job_state
+
+
+def handle_shieldhit_cancellation(tasks: list[CeleryTaskModel]):
+    """Function cancelling shieldhit processes"""
+    command_as_list = ['kill', '--signal', 'SIGINT']
```
grzanka: what about `shieldhit_kill_many_cmd` ?

yaptide/celery/utils/manage_tasks.py
```diff
@@ -87,3 +92,54 @@ def translate_celery_state_naming(job_state: str) -> str:
...
+    celery_ids = []
```
grzanka: what about `task_ids_to_cancel`

yaptide/celery/utils/manage_tasks.py
```diff
@@ -87,3 +92,54 @@ def translate_celery_state_naming(job_state: str) -> str:
...
+    for task in tasks:
+        if task.task_state == EntityState.RUNNING.value:
+            command_as_list.append(str(task.sim_pid))
+        elif task.task_state in (EntityState.PENDING.value, EntityState.UNKNOWN.value) or task.sim_pid is None:
+            celery_ids.append(task.celery_id)
+
+    # Send SIGINT to Shieldhit processes
```
grzanka:
```suggestion
    # Send SIGINT to shieldhit processes, that will immediately stop the simulation and trigger output files dump
```

yaptide/celery/utils/manage_tasks.py
```diff
@@ -87,3 +92,54 @@ def translate_celery_state_naming(job_state: str) -> str:
...
+    subprocess.run(command_as_list, check=True)
+    celery_app.control.revoke(celery_ids, terminate=True, signal="SIGINT")
```
grzanka:
```suggestion
    # tell celery to stop all other tasks related to this simulation, being in pending or unknown status
    celery_app.control.revoke(celery_ids, terminate=True, signal="SIGINT")
```

yaptide/celery/utils/manage_tasks.py
```diff
@@ -87,3 +92,54 @@ def translate_celery_state_naming(job_state: str) -> str:
         return EntityState.COMPLETED.value
     # Others are the same
     return job_state
+
+
+def handle_shieldhit_cancellation(tasks: list[CeleryTaskModel]):
+    """Function cancelling shieldhit processes"""
```
grzanka:
```suggestion
    """Function triggering shieldhit processes to terminate and dump results"""
```

yaptide/celery/utils/manage_tasks.py
```diff
@@ -87,3 +92,54 @@ def translate_celery_state_naming(job_state: str) -> str:
...
+        elif task.task_state in (EntityState.PENDING.value, EntityState.UNKNOWN.value) or task.sim_pid is None:
+            celery_ids.append(task.celery_id)
+
+    # Send SIGINT to Shieldhit processes
+    subprocess.run(command_as_list, check=True)
+    celery_app.control.revoke(celery_ids, terminate=True, signal="SIGINT")
+
+
+def handle_fluka_cancellation(tasks: list[CeleryTaskModel]):
+    """Function cancelling fluka processes"""
```
grzanka:
```suggestion
    """Function cancelling fluka processes with results dump"""
```

yaptide/celery/utils/manage_tasks.py
```diff
@@ -87,3 +92,54 @@ def translate_celery_state_naming(job_state: str) -> str:
...
+    for task in tasks:
+        if task.task_state == EntityState.RUNNING.value:
+            try:
+                if (target_folder := next(Path(task.path_to_sim).glob("fluka_*/"), None)):
+                    (target_folder / FILE_NAME).write_text("")
+            except OSError as e:
+                logging.warning("Error due to file operation: %s", str(e))
+        elif task.task_state in (EntityState.PENDING.value, EntityState.UNKNOWN.value) or task.path_to_sim is None:
+            celery_ids.append(task.celery_id)
+    celery_app.control.revoke(celery_ids, terminate=True, signal="SIGINT")
```
grzanka: maybe you should wait till results are really written to the disk before terminating all Fluka tasks ?
if you terminate task while in the process of writing the output you will get corrupted file.
if I remember correctly Fluka would notify you on writing the results by removing the `rfluka.stop` file.

yaptide/celery/utils/manage_tasks.py
```diff
@@ -87,3 +92,54 @@ def translate_celery_state_naming(job_state: str) -> str:
...
+def cancel_tasks_without_fetching(simulation: CelerySimulationModel, tasks: list[CeleryTaskModel]):
+    """Function to cancel tasks without feching data"""
+    celery_ids = [
+        task.celery_id for task in tasks
+        if task.task_state in [EntityState.PENDING.value, EntityState.RUNNING.value, EntityState.UNKNOWN.value]
+    ]
+
+    # Revoke the merge task first
+    celery_app.control.revoke(simulation.merge_id, terminate=True, signal="SIGINT")
+    celery_app.control.revoke(celery_ids, terminate=True, signal="SIGINT")
```
grzanka, line 125: is that blocking or non-blocking operation ?

yaptide/celery/utils/manage_tasks.py
NiEkOKsU: Task ids are appended to a command_as_list

yaptide/celery/utils/manage_tasks.py
grzanka: i.e. after calling `revoke` are you 100% sure that all fluka processes are dead ? or you need to wait till its done ?

NiEkOKsU: non-blocking

yaptide/celery/utils/progress/fluka_monitor.py
```diff
@@ -153,3 +155,31 @@ def read_fluka_out_file(event: threading.Event,
...
+                if line.startswith(S_TERMINATED_FROM_OUTSIDE) or line.startswith(S_OK_OUT_COLLECTED):
+                    break
+                if in_progress:
+                    simulated_primaries, remaining_particles = parse_progress_remaining_line(line)
+                    if requested_primaries == 0:
+                        requested_primaries = simulated_primaries + remaining_particles
+                    in_progress = False
+                else:
+                    if line.startswith(S_OK_OUT_IN_PROGRESS):
+                        in_progress = True
```
grzanka: can you explain the logic here in some comment lines ?
I don't get why do you check for requested == 0 ? and then calculate `requested_primaries = simulated_primaries + remaining_particles`

yaptide/celery/utils/pymc.py
```diff
@@ -169,14 +175,18 @@ def get_fluka_estimators(dir_path: Path) -> dict:
     return estimators_dict
 
 
-def average_values(base_values: List[float], new_values: List[float], count: int) -> List[float]:
+def average_values(base_values: List[float], new_values: List[float], base_particles: int,
```
grzanka, line 215: that function needs at least 20 lines of comments with good explanation and some examples.

yaptide/persistence/models.py
```diff
@@ -269,6 +269,12 @@ class CeleryTaskModel(TaskModel):
     __tablename__ = 'CeleryTask'
     id: Column[int] = db.Column(db.Integer, db.ForeignKey('Task.id', ondelete="CASCADE"), primary_key=True)
     celery_id: Column[str] = db.Column(db.String, nullable=False, default="", doc="Celery task ID")
+    sim_pid: Column[int] = db.Column(db.Integer,
+                                     nullable=True,
+                                     doc='Id of simulation process used to communicate with shieldhit process')
```
grzanka: we typically speak of PID, that would clearly differentiate it from celery IDs.

grzanka: does it mean you introduce a column which is used only for SHIELD-HIT12A ? if that is the case, write in doc what is its meaning for Fluka ?

yaptide/persistence/models.py
```diff
@@ -269,6 +269,12 @@ class CeleryTaskModel(TaskModel):
...
+    path_to_sim: Column[str] = db.Column(db.String,
```
grzanka: maybe better to call that directory a workspace ?

grzanka: also - why do you keep that in the DB ? such directory is usually removed once the job is done

yaptide/celery/utils/manage_tasks.py
NiEkOKsU: It terminates only workers that don't start simulation before user click cancel button. Descripted case don't occurs.

yaptide/persistence/models.py
NiEkOKsU: Celery id is a id of a celery worker, sim_pid is a PID of a shieldhit process

yaptide/celery/utils/progress/fluka_monitor.py
NiEkOKsU: It is the way to calculate a requestred parameter value, because fluka out file doesn't contain a info about it inside of it. We have got only a simulated primaries and remaining, it is sufficient to calculate requested primaries. I want count it only once, so for this reason there is == 0.

yaptide/persistence/models.py
NiEkOKsU: If it could prove aby problems in future I can remove path after the end of the job. This is in a dataabse because in a Flask backend the dataabse is a only option  to get some information about the task.

yaptide/celery/utils/progress/fluka_monitor.py
grzanka: what you write isn't very obvious. It's the reason why you've written in such a way it's written. The reasons why the code is written such way should go in the comment. That's main purpose of comments

yaptide/celery/utils/manage_tasks.py
grzanka: I just suggested variable rename, `command_as _list` is not so meaningful.

yaptide/celery/utils/manage_tasks.py
grzanka: OK, I've got it now. So indeed this requires very good comments.
You needs to say explicitely that you ask Fluka to dump results and stop by creating `rfluka.stop` file. The comment should say that you deliberately DO NOT revoke such tasks, cause you need to wait till it dump results to disk (which may take some time, couple of seconds), and you expect that this tasks will exit by itself.

yaptide/celery/utils/manage_tasks.py
grzanka: that was suggestion to rename variable

yaptide/celery/utils/manage_tasks.py
grzanka: then I doubt if you should immediately set the simulation state to CANCELED if you are not sure that it has already been killed. Killing may take some time. As for now its OK, but I think this should be done in better way later.

yaptide/batch/batch_methods.py
```diff
@@ -327,3 +327,31 @@ def delete_job(simulation: BatchSimulationModel, user: KeycloakUserModel,
         return {"message": "Job cancelation failed"}, 500
 
     return {"message": "Job canceled"}, 200
+
+
+def cancel_simulation_with_fetching_data(simulation: BatchSimulationModel, user: KeycloakUserModel,
+                                         cluster: ClusterModel) -> tuple[dict, int]:  # skipcq: PYL-W0613
```
grzanka:
```suggestion
                                         cluster: ClusterModel) -> tuple[dict, int]:  # skipcq: PYL-W0613
```

why do you need `# skipcq: PYL-W0613` ?

yaptide/batch/batch_methods.py
```diff
@@ -327,3 +327,31 @@ def delete_job(simulation: BatchSimulationModel, user: KeycloakUserModel,
...
+    """Function that stops simulation without cancelling collection job"""
+    array_id = simulation.array_id
+    tasks = fetch_batch_tasks_by_sim_id(sim_id=simulation.id)
+    try:
+        con = get_connection(user=user, cluster=cluster)
+        i = 1
```
grzanka, line 339: why do you need `i` here ?

grzanka: I do not see i being incremented...

yaptide/celery/tasks.py
```diff
@@ -257,10 +270,16 @@ def merge_results(results: list[dict]) -> dict:
 
         if averaged_estimators is None:
             averaged_estimators: list[dict] = result.get("estimators", [])
+            total_primaries += int(averaged_estimators[0]["metadata"]["number_of_primaries"])
+            logging.debug(total_primaries)
             # There is nothing to average yet
             continue
 
-        averaged_estimators = average_estimators(averaged_estimators, result.get("estimators", []), i)
+        new_particles = int(result.get("estimators", [])[0]["metadata"]["number_of_primaries"])
```
grzanka: w have primaries, not particles when talking about number of simulated items

yaptide/celery/utils/pymc.py
```diff
@@ -112,23 +112,29 @@ def __call__(self, file_path: str, rng_seed: int) -> None:
...
-                                           text=True)
-        logging.info("simulation subprocess with return code %d finished", completed_process.returncode)
-
+        process = subprocess.Popen(command_as_list,
+                                   cwd=str(dir_path),
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.PIPE,
+                                   text=True)
+        pid = process.pid
+        logging.info("sending info to task: %d in simulation: %d with update key: %s to update PID %d", task_id,
```
grzanka: why `simulation_id or -1` feeded into `%d` ??

grzanka:
Tested on docker and it badly fails with:

```
2024-12-15 17:17:26,532] ERROR in app: Exception on /jobs/direct [DELETE]
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/flask/app.py", line 902, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/flask_restful/__init__.py", line 489, in wrapper
    resp = resource(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/flask/views.py", line 110, in view
    return current_app.ensure_sync(self.dispatch_request)(**kwargs)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/flask_restful/__init__.py", line 604, in dispatch_request
    resp = meth(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/app/yaptide/routes/utils/decorators.py", line 24, in wrapper
    return f(user, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/app/yaptide/routes/celery_routes.py", line 170, in delete
    handle_shieldhit_cancellation(tasks)
  File "/usr/local/app/yaptide/celery/utils/manage_tasks.py", line 109, in handle_shieldhit_cancellation
    subprocess.run(command_as_list, check=True)
  File "/usr/local/lib/python3.12/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/local/lib/python3.12/subprocess.py", line 1955, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'kill'
ERROR:yaptide:Exception on /jobs/direct [DELETE]
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/flask/app.py", line 902, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/flask_restful/__init__.py", line 489, in wrapper
    resp = resource(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/flask/views.py", line 110, in view
    return current_app.ensure_sync(self.dispatch_request)(**kwargs)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/flask_restful/__init__.py", line 604, in dispatch_request
    resp = meth(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/app/yaptide/routes/utils/decorators.py", line 24, in wrapper
    return f(user, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/app/yaptide/routes/celery_routes.py", line 170, in delete
    handle_shieldhit_cancellation(tasks)
  File "/usr/local/app/yaptide/celery/utils/manage_tasks.py", line 109, in handle_shieldhit_cancellation
    subprocess.run(command_as_list, check=True)
  File "/usr/local/lib/python3.12/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/local/lib/python3.12/subprocess.py", line 1955, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'kill'
```

grzanka:
It seems there are some issues to run kill on`python:3.12-slim` image

grzanka:
Maybe better to use `os.kill(pid, signal.SIGTERM)` ?

yaptide/celery/utils/manage_tasks.py
```diff
@@ -87,3 +92,58 @@ def translate_celery_state_naming(job_state: str) -> str:
...
+    celery_ids = []
+    for task in tasks:
+        if task.task_state == EntityState.RUNNING.value:
+            logging.warning(task.sim_pid)
+            command_as_list.append(str(task.sim_pid))
+        elif task.task_state in (EntityState.PENDING.value, EntityState.UNKNOWN.value) or task.sim_pid is None:
+            celery_ids.append(task.celery_id)
+
+    # Send SIGINT to shieldhit processes, that will immediately stop the simulation and trigger output files dump
+    subprocess.run(command_as_list, check=True)
```
grzanka: it seems better option is to use `os.kill(pid, signal.SIGTERM)` as `kill` may not be available on docker python image

grzanka: There is another thing - what if we deploy yaptide in such way that there are couple of celery workers, running on different VMs and communicating by network via redis. 
Then some of the parallel tasks would run on machine A and some on machine B.
Flask itself would run on machine C.

There is no way to kill a process on machine B, being on machine C.

yaptide/celery/utils/pymc.py
NiEkOKsU: simulation_id is set to None is tests (thts why some tests failing before, they failed on this log). I will wroite it in comment

yaptide/batch/batch_methods.py
NiEkOKsU: It is used in a loop under, but i don't remember to add a incrementation of this in that loop.

yaptide/batch/batch_methods.py
NiEkOKsU: I copied the above function, I didn't know it had any use (in fact I didn't even think about it)

yaptide/batch/batch_methods.py
grzanka: maybe an `for i, task in enumerate(tasks)` would do the job as well ?

yaptide/celery/utils/pymc.py
grzanka: do you know why `simulation_id` is set to none ? that seems like sign of some bug.
Hiding bug in complex logic is a bad habit, better to report it in an issue.
In the issue you can write that you suspect sim_id being None and you should say how to reproduce it.

In the code here I would rather use somehow more verbose code, like:

```
if simulation_id:
   logging.info(...)
   send_task_update(...)
else:
  # we may have simulation_id being none, reported in an issue ISSUE_URL...
  logging.info(....)
```

does it make sense to update (send task update) a task which doesn't belong to any simulation ? do we have such tasks ?

yaptide/batch/batch_methods.py
grzanka: The please get rid of `# skipcq: PYL-W0613` and see if deepsource complains

yaptide/celery/utils/pymc.py
NiEkOKsU: test_celery is running without a [simulatiion_id](tests/integration/test_celery.py#L60-66) in a some reason. In some reason too we are enabling [running a simulation](tyaptide/celery/tasks.py#L29-35) without giving a information about a simulation_id

grzanka:
image

yaptide/celery/utils/pymc.py
grzanka: Then please write it down in an issue and put here the link in the coment

yaptide/celery/utils/manage_tasks.py
```diff
@@ -96,40 +92,30 @@ def translate_celery_state_naming(job_state: str) -> str:
...
-        elif task.task_state in (EntityState.PENDING.value, EntityState.UNKNOWN.value) or task.path_to_sim is None:
-            celery_ids.append(task.celery_id)
-
-    # tell celery to stop all other tasks related to this simulation, being in pending or unknown status
-    celery_app.control.revoke(celery_ids, terminate=True, signal="SIGINT")
+    # change celery tasks state to brake a while loop and dump data
+    if len(celery_ids_to_dump_data) > 0:
+        for id in celery_ids_to_dump_data:
+            result = AsyncResult(id)
+            result.backend.store_result(id, None, 'DUMP')
```
grzanka: Instead of custom celery state I would prefer to use metadata approach, like described here:

```
@app.task(bind=True)
def long_running_task(self):
    print("Starting task...")
    for i in range(10):
        time.sleep(1)
        print(f"Processing step {i+1}/10...")

        # Check task metadata for a "dump" command
        task_info = self.backend.get_task_meta(self.request.id)
        if task_info.get("dump"):
            print("Dump signal received. Dumping data...")
            # Perform dump logic here
            self.backend.store_result(self.request.id, {"dump": False}, state="RUNNING")

    print("Task completed.")

# Sending a "dump" signal
def send_dump_signal(task_id):
    result = AsyncResult(task_id)
    result.backend.store_result(task_id, {"dump": True}, state="RUNNING")

# Example usage
task = long_running_task.apply_async()
send_dump_signal(task.id)
```
It's quite dangerous to introduce own states...

grzanka: Check which backends support metadata, redis and DB should be OK, rabbitmq probably not...

grzanka: another solution, which may be more generic - secured message queue on the backend server, to which all workers: celery and ares subscribe. Once they read a message to dump something they dump it. It may be more flexible solution, as you could parametrize it, i.e. dump only particular Page or Estimator.

NiEkOKsU: Why do you think a custom states could be dangerous?

grzanka: I think we can stick with something based on `self.backend.store_result` now and think of more generic solution in the next step.

grzanka: > Why do you think a custom states could be dangerous?

For couple of reasons:
1. Consistency: we already have simulation and backend states (and their transitions) defined in the backend model. Introducing more states which are inconsistent from what we have a model is an issue.
2. State transitions in celery. Celery itself knows about transitions between built-in states and uses that information in many places. See https://docs.celeryq.dev/en/stable/userguide/tasks.html#task-states and https://docs.celeryq.dev/en/stable/reference/celery.states.html. Introduction a state with unknow transition may lead to undefined behaviour.
3. Complexity. Introducing new state increases complexity of the state model.

yaptide/celery/utils/manage_tasks.py
```diff
@@ -110,12 +110,11 @@ def handle_cancellation_with_fetching(tasks: list[CeleryTaskModel]):
     if len(celery_ids_to_dump_data) > 0:
         for id in celery_ids_to_dump_data:
             result = AsyncResult(id)
-            result.backend.store_result(id, None, 'DUMP')
+            result.backend.store_result(id, {"dump": True}, state="RUNNING")
```
grzanka: how would that work if in the meantime task would go from running to completed state.Do you know if running that command will bring it back to running state ?

in this function (`handle_cancellation_with_fetching`) you fetch task state from the database, not from celery. Updates from celery to the DB are not so often, so it may happen as well that the celery task state at given moment of time is different that the task state in the DB.

yaptide/celery/utils/pymc.py
```diff
@@ -112,23 +115,46 @@ def __call__(self, file_path: str, rng_seed: int) -> None:
...
-        command_stdout = completed_process.stdout
-        command_stderr = completed_process.stderr
+        process = subprocess.Popen(command_as_list,
+                                   cwd=str(dir_path),
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.PIPE,
+                                   text=True)
+        try:
+            # wait untill simulation subprocess is not marked in celery as ended
+            while not state.get("end", False):
```
grzanka: are we 100% sure that this loop will never become endless ?
there is no mechanism to protect it from endless working ?
one may wish to kill that loop once simultion process is over

yaptide/celery/utils/pymc.py
```diff
@@ -112,23 +115,46 @@ def __call__(self, file_path: str, rng_seed: int) -> None:
...
+                time.sleep(1)
+                # Check if the task has been marked to dump
+                simulation = AsyncResult(celery_id)
```
grzanka: do you need to get simulation object inside the loop ? maybe it would be enough to get it once outside the loop and just call `get_task_meta` inside.

yaptide/celery/utils/pymc.py
```diff
@@ -112,23 +115,46 @@ def __call__(self, file_path: str, rng_seed: int) -> None:
...
-        command_stderr = completed_process.stderr
+        process = subprocess.Popen(command_as_list,
+                                   cwd=str(dir_path),
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.PIPE,
+                                   text=True)
+        try:
+            # wait untill simulation subprocess is not marked in celery as ended
+            while not state.get("end", False):
+                time.sleep(1)
```
grzanka: that sleep time should be parametrized with reasonable default (1 second is ok)

yaptide/celery/utils/pymc.py
```diff
@@ -112,23 +115,46 @@ def __call__(self, file_path: str, rng_seed: int) -> None:
...
+                # Check if the task has been marked to dump
+                simulation = AsyncResult(celery_id)
+                state = simulation.backend.get_task_meta(celery_id)
+                if state.get("dump", False):
```
grzanka: Can this be called with named parameters, like `state.get(name="dump", default=False)` ?

yaptide/celery/utils/pymc.py
```diff
@@ -294,13 +325,17 @@ def read_file(event: threading.Event,
     }
     logging.info("Sending final update for task %d, simulated primaries %d", task_id, simulated_primaries)
     send_task_update(simulation_id, task_id, update_key, up_dict)
+    result = AsyncResult(celery_id)
+    # mark celery task as ended to end a while loop waiting for results
+    result.backend.store_result(celery_id, {"end": True}, state="RUNNING")
```
grzanka: thats so much unobvious - why do want to kill the task when you are done with reading logfile from log monitoring function. 

I would rather rely on seeing what happens with the process you've executed. If that process terminates then its good moment to stop the "wait for dump message" loop.

Anyway many things may still go one after simulation binary process is done. You may still like to work on the output, send task updates, digest the logs etc...

The "wait for dump signal" loop should be active only during lifetime of simulation process.

grzanka:
@NiEkOKsU take a look at failing tests, i.e.  `tests/integration/test_run_crashing_simulation.py::test_run_simulation_with_flask_crashing `

PR #1263, opened 2024-12-28
1150 create template scripts for running simulations using fluka

dudiiiiiiii:


grzanka:
@dudiiiiiiii may this one be close as there is new one here: https://github.com/yaptide/yaptide/pull/1278 ?

dudiiiiiiii:
> @dudiiiiiiii may this one be close as there is new one here: #1278 ?

Yes, i wanted to test te second one on yap-dev but flask container has some problems again

PR #1432, opened 2025-06-12
Add ELK to visualize and store logs

KacperST:
# What was done and is working:
- ELK can be opened and logs can be seen (but probably we need to add a config file to setup ELK when container is first created)
- Added healtcheck for ELK
image

# What is not working:
- We didnt figure out yet how to add logs from flask to ELK - something is not working
- We did not test if volume works


yaptide/application.py
```diff
@@ -12,12 +14,20 @@ def create_app():
     """Function starting Flask Server"""
     flask_name = __name__.split('.')[0]
     app = Flask(flask_name)
+    
+    logstash_host = "logstash"
+    logstash_port = 5001
+    logstash_handler = logstash.TCPLogstashHandler(logstash_host, logstash_port, version=1)
+    logstash_handler.setLevel(logging.DEBUG)
```
grzanka: logging level should be parametrized somehow and controlled for example by env variables. It's not the best idea to have DEBUG level on production.

docker-compose.yml
```diff
@@ -31,6 +31,8 @@ services:
     image: yaptide_flask
     container_name: yaptide_flask
     restart: unless-stopped
+    ports:
```
grzanka: is that really needed ? API calls on production should go via nginx, port 6000 is only used for internal communication. 

We should only expose what is really needed. Is it needed to expose port  6000 ?

logstash.conf
```diff
@@ -0,0 +1,14 @@
...
+    codec => json
+  }
+}
+
+output {
+  elasticsearch {
+    hosts => ["elasticsearch:9200"]
+    index => "yaptide-logs-%{+YYYY.MM.dd}"
+  }
+  stdout { codec => rubydebug }
```
grzanka, line 13: why rubydebug ?

yaptide/application.py
```diff
@@ -12,12 +14,20 @@ def create_app():
     """Function starting Flask Server"""
     flask_name = __name__.split('.')[0]
     app = Flask(flask_name)
+    
+    logstash_host = "logstash"
+    logstash_port = 5001
```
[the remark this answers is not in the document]
grzanka: I fully agree with Copilot

docker-compose.yml
```diff
@@ -117,7 +119,57 @@ services:
       yaptide_helper_worker:
         condition: service_healthy
 
+  logstash:
```
grzanka, line 125: Consider moving the whole logging stuff into a separate docker compose. You can then chain the compose files in following way:

```
docker compose --file docker-compose.yml --file docker-compose-develop.yml --file docker-compose.fast.yml up  --detach
```

In this way default config would be with stdout logging. ELK would be optional and deployed on complex dev env and in production.

grzanka: See https://github.com/yaptide/yaptide/blob/feature/add_elk/scripts/start_with_docker_develop.sh

docker-compose.yml
KacperST: Nope it's not needed. I must have mistakenly forgotten to remove that after debugging

grzanka:
Resolve conflicts:

image

yaptide/application.py
```diff
@@ -12,12 +15,31 @@ def create_app():
     """Function starting Flask Server"""
     flask_name = __name__.split('.')[0]
     app = Flask(flask_name)
+
```
grzanka, line 18: Do we really need that ?

Can't we simply add a logging driver in docker compose ?
https://docs.docker.com/engine/logging/configure/

Then in principle no changes should be needed to the python code.

```
    logging:
      driver: "gelf"
      options:
        gelf-address: "udp://localhost:5000"
```

grzanka: Check https://github.com/RADMONQA/radiation_monitors_app/pull/21/files

Glonk223: I'll check it out thank you!

grzanka:
We can reopen if the work resumes or some new volunteers come

PR #1502, opened 2025-09-01
Allow for filtering jobs by job_state

kmichalikk:
This PR adds `job_state` query param with validation for fetching simulations which allows to select simulations with subset of states. This is required 1) to filter out UNKNOWN simulations which front end is not able to display 2) add "Last 5 simulations" section with only PENDING, RUNNING, MERGING, COMPLETED states

### DIFF
--- a/yaptide/routes/user_routes.py
+++ b/yaptide/routes/user_routes.py
@@ -1,7 +1,8 @@
 import logging
 from enum import Enum
+from typing import List
 
-from flask import request, current_app as app
+from flask import request
 from flask_restful import Resource
 from marshmallow import Schema, fields, ValidationError
 from sqlalchemy import asc, desc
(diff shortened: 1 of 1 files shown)


yaptide/routes/user_routes.py
```diff
@@ -30,6 +30,24 @@ class OrderBy(Enum):
     END_TIME = "end_time"
 
 
+def validate_job_state(states):
```
grzanka: add some type hinting

yaptide/routes/user_routes.py
```diff
@@ -30,6 +30,24 @@ class OrderBy(Enum):
...
+    """check whether the list of job states in query contains actual enum values"""
+    app.logger.error(f'states {states}')
+    valid_states = [es.value for es in EntityState]
+    for state in states:
+        if state not in valid_states:
+            raise ValidationError('Invalid job state')
```
grzanka: what about?

```
# check if all states belong to EntityStates
if not set(states).issubset({es.value for es in EntityState})
```

grzanka:
@kmichalikk fix deepsource complains

grzanka (approved):

Issue #1568, opened 2025-11-08
Merging jobs on celery endpoint is very slow


grzanka:
I've started some simple simulation with SHIELD-HIT12A on celery backend.
You can see the progress here:

https://github.com/user-attachments/assets/e918c39d-86af-408c-8433-7f1a3ab0b48e

The job was submited and entered running state at 00:14, it was in running state at 00:16
The progress bar moved completely to the right at 00:26 , which was 13 seconds into running simulation.

<img width="368" height="443" alt="Image" src="https://github.com/user-attachments/assets/4847439a-7fc8-41cf-be0e-b01728274960" />

Then for next 7-8 seconds job was in the RUNNING state, while the progress bar was not moving and was still at 100%.
It entered MERGING_QUEUED state at 00:34.

What was happening during that time ? All the 14 tasks should be completed at 00:26 and the jobs should enter MERGING_QUEUED state immediately.

By looking at the HAR file we see the request to https://yap-dev.c3.plgrid.pl:8443/jobs/direct at 17:30:26

<img width="1223" height="212" alt="Image" src="https://github.com/user-attachments/assets/e8a6ce00-27bc-4870-8eed-0a0cfffa58e3" />

Two seconds later at 17:30:28 we see a query to check job status and all the tasks are in running state:

<img width="1233" height="677" alt="Image" src="https://github.com/user-attachments/assets/a188f540-4136-4b3e-9c30-fe914df4ba9b" />

```json
{
  "job_state": "RUNNING",
  "job_tasks_status": [
    {
      "last_update_time": "Sat, 08 Nov 2025 16:30:28 GMT",
      "requested_primaries": 714,
      "simulated_primaries": 1,
      "start_time": "Sat, 08 Nov 2025 16:30:27 GMT",
      "task_id": 2790,
      "task_state": "RUNNING"
    },
    {
      "last_update_time": "Sat, 08 Nov 2025 16:30:27 GMT",
      "requested_primaries": 714,
      "simulated_primaries": 0,
      "start_time": "Sat, 08 Nov 2025 16:30:27 GMT",
      "task_id": 2791,
      "task_state": "RUNNING"
    },
    {
      "last_update_time": "Sat, 08 Nov 2025 16:30:27 GMT",
      "requested_primaries": 714,
      "simulated_primaries": 0,
      "start_time": "Sat, 08 Nov 2025 16:30:27 GMT",
      "task_id": 2792,
      "task_state": "RUNNING"
    },
    {
[...]

PR #1599, opened 2025-12-08
Investigate merging performance

lythx:
Related to the https://github.com/yaptide/yaptide/issues/1568

### DIFF
--- a/yaptide/batch/watcher.py
+++ b/yaptide/batch/watcher.py
@@ -10,6 +10,7 @@ from datetime import datetime
 from io import TextIOWrapper
 from pathlib import Path
 from urllib import request
+import yaptide.performance_tracker as tracker
 
 RUN_MATCH = r"\bPrimary particle no.\s*\d*\s*ETR:\s*\d*\s*hour.*\d*\s*minute.*\d*\s*second.*\b"
 COMPLETE_MATCH = r"\bRun time:\s*\d*\s*hour.*\d*\s*minute.*\d*\s*second.*\b"
@@ -25,6 +26,7 @@ def log_generator(thefile: TextIOWrapper, event: threading.Event = None, timeout
     Main purpose is monitoring of the log files
     """
     sleep_counter = 0
+    line_wait_id = None
     while True:
         if event and event.is_set():
             break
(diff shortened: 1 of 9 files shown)


yaptide/celery/utils/requests.py
```diff
@@ -2,10 +2,12 @@
 import os
 
 import requests
+import yaptide.performance_tracker as tracker
```
grzanka, line 8: it's not the best idea to use shared module level object, which is then used in code executed in different instances, processes or threads. I would rather go for appending to process/thread/core-specific file or simply print to the stdout and the grep thought the logs.

Shared object without proper ownership (mutexes etc) may lead to confused results.

yaptide/batch/watcher.py
```diff
@@ -38,12 +38,17 @@ def log_generator(thefile: TextIOWrapper, event: threading.Event = None, timeout
             if line_wait_id is None:
                 line_wait_id = tracker.start("log_generator line wait")
 
-            time.sleep(1)
-            sleep_counter += 1
+            wait_interval = 0.1
```
grzanka, line 41: this generator somehow polls the file, so maybe the better name should be `polling_interval_s` ? As such it should go to parameter of the `log_generator` function with default of 0.1

yaptide/celery/utils/progress/fluka_monitor.py
```diff
@@ -87,7 +87,7 @@ def check_progress(line: str, next_backend_update_time: int, details: TaskDetail
                 "start_time": utc_without_offset(progress_details.utc_now),
                 "task_state": EntityState.RUNNING.value
             }
-            send_task_update(details.simulation_id, details.task_id, details.update_key, up_dict)
+            send_task_update(details.simulation_id, details.task_id, details.update_key, up_dict, async_send=True)
```
grzanka, line 90: async call would be faster, but do we have guarantee that the proper sequence of updates and state transistions will be preserved ? What if all that updates ends up in a queue to execute and are still being fired, even if `shieldhit` process finished working long ago ?

yaptide/celery/utils/requests.py
```diff
@@ -1,17 +1,63 @@
...
 
+_THREAD_LOCAL = threading.local()
+_REQUEST_TIMEOUT = (3.0, 10.0)
+_EXECUTOR = ThreadPoolExecutor(max_workers=1)
 
-def send_task_update(simulation_id: int, task_id: int, update_key: str, update_dict: dict) -> bool:
-    """Sends task status to backend which will update the database"""
-    id = tracker.start("send_task_update")
+
+def _get_session() -> requests.Session:
```
grzanka, line 15: intstead of separate thread, you can use as well dedicated `celery` queue to fire long running REST requests:
https://github.com/yaptide/yaptide/blob/master/Dockerfile-helper-worker

PR #1617, opened 2026-01-24
Remove sleep delays and standarize time-related variables

lythx:
`time.sleep()` calls were causing unnecessary delay on simulation task end.

- Replaced `time.sleep()` with `event.wait()` in functions related to task monitoring. When task ends, `event.set()` gets called so `event.wait()` exits immediately. 

- Changed names of time-related parameters to be more descriptive and added `_unit` suffix to avoid confusion (e.g. `max_idle_seconds`, `polling_interval_seconds`). In some cases the values might be lower than 1 second (e.g. polling intervals for small simulations), so I also changed the types to `float` instead of `int` (changed it for all time variables for consistency).

- Made all timeouts/intervals configurable in `read_file()` functions. Added descriptions for these parameters in docstrings to explain what each of them does in detail.

- Changed `read_file()` to `read_shieldhit_file()` for clarity (consistent with `read_fluka_file()` function).

- Changed the return type of `log_generator` from `str` to `Iterator[str]`. `Iterator` is from `collections.abc` package, which is in the standard library, so it should be fine on Ares HPC.

I ran some simulations on my pc and on the yap-dev instance to test this. Everything seems to be working as intended.

### DIFF
--- a/yaptide/batch/watcher.py
+++ b/yaptide/batch/watcher.py
@@ -19,11 +19,10 @@ REQUESTED_MATCH = r"\bRequested number of primaries NSTAT"
 TIMEOUT_MATCH = r"\bTimeout occured"
 
 
-def log_generator(
-        thefile: TextIOWrapper, 
-        event: threading.Event = None, 
-        max_idle_seconds: float = 3600, 
-        polling_interval_seconds: float = 1) -> Iterator[str]:
+def log_generator(thefile: TextIOWrapper,
+                  event: threading.Event = None,
+                  max_idle_seconds: float = 3600,
+                  polling_interval_seconds: float = 1) -> Iterator[str]:
     """
     Generator equivalent to `tail -f` Linux command.
     Yields new lines appended to the end of the file.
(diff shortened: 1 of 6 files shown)


lythx:
`DeepSource: Python` check failed for `read_shieldhit_file` function complexity. I guess it's because using event.wait() adds an if statement or maybe the function wasn't passing the test before, but I renamed it so it got re-checked.

yaptide/batch/watcher.py
```diff
@@ -68,17 +80,37 @@ def send_task_update(sim_id: int, task_id: str, update_key: str, update_dict: di
...
+                        update_key: str,
+                        backend_url: str,
+                        max_wait_for_file_seconds: float = 30,
+                        max_idle_seconds: float = 3600,
+                        update_interval_seconds: float = 2,
+                        polling_interval_seconds: float = 1):  # skipcq: PYL-W0613
+    """
+    Monitors log file of a shieldhit task and sends updates to the backend.
+
+    Args:
```
grzanka, line 99: if you want to add args description (which is nice) be consistent and cover all of them, not only some subset

yaptide/batch/watcher.py
```diff
@@ -68,17 +80,37 @@ def send_task_update(sim_id: int, task_id: str, update_key: str, update_dict: di
...
+                        sim_id: int,
+                        task_id: int,
+                        update_key: str,
+                        backend_url: str,
+                        max_wait_for_file_seconds: float = 30,
+                        max_idle_seconds: float = 3600,
+                        update_interval_seconds: float = 2,
+                        polling_interval_seconds: float = 1):  # skipcq: PYL-W0613
+    """
+    Monitors log file of a shieldhit task and sends updates to the backend.
```
grzanka, line 94: You can mention that main purpose of the updates is the progress bar update and state updates (like simulation failed or completed)

grzanka:
> `DeepSource: Python` check failed for `read_shieldhit_file` function complexity. I guess it's because using event.wait() adds an if statement or maybe the function wasn't passing the test before, but I renamed it so it got re-checked.

You can silence it with skipcq for [PY-R1000](https://deepsource.com/directory/python/issues/PY-R1000)

grzanka (approved):

PR #1630, opened 2026-03-16
1618 - simulations with more tasks than primaries fail

### LINKED ISSUE #1618: Simulations with more tasks than primaries don't work
The number of primaries per simulation has to be the same for each task, so if there are more tasks than primaries it gets rounded down to 0 and the simulation breaks. This can be fixed either by setting tasks=primaries in the backend if the user inputs tasks>primaries, or by rejecting such simulations.

Mosder:
This PR adds a quick and easy fix of the bug mentioned in #1618 by taking a minimum of tasks and primaries.
I do think it may be kind of weird to see the task count be different from the requested one, so I believe that it maybe should be blocked on the inputs themselves, on the front-end side. Though the fix should probably still be included if someone wants to run simulations directly through the API calls.

I'm also interested by the phenomenon that number of primaries is also able to go down from the inputted one when it's value is not a divisor of number of tasks as it's being divided between them with no remainder (e.g. 19 primaries and 10 tasks => 1 primary per task => 10 primaries total). I presume it's not that big of a deal, since you usually want to have the number of primaries magnitudes greater than the number of tasks, but it's something I can change if it's worth pursuing.

Description:

This pull request introduces several improvements and refactorings to how simulation input types and the number of simulation tasks (`ntasks`) are handled and validated. The changes unify and clarify the use of input type enums, add robust validation for `ntasks` to prevent invalid simulation configurations, and refactor functions to be more general and maintainable.

Key improvements and changes:

**Input type handling and enums:**
* Replaced the custom `JSON_TYPE` enum with the standardized `InputType` enum throughout the codebase, updating all references and related logic for determining input types. [[1]](diffhunk://#diff-fd6ef04bceceed71a763cfa5341d99a30fd6c65292e0ac6922449b237079b25dL39-R151) [[2]](diffhunk://#diff-fd6ef04bceceed71a763cfa5341d99a30fd6c65292e0ac6922449b237079b25dL70-R171) [[3]](diffhunk://#diff-fd6ef04bceceed71a763cfa5341d99a30fd6c65292e0ac6922449b237079b25dL177-R292) [[4]](diffhunk://#diff-8670b995eb9f58b14f5a7cab50215edce3442835f8cb4d7470ff657caf7a3820R5-R8) [[5]](diffhunk://#diff-8670b995eb9f58b14f5a7cab50215edce3442835f8cb4d7470ff657caf7a3820L81-R83) [[6]](diffhunk://#diff-8670b995eb9f58b14f5a7cab50215edce3442835f8cb4d7470ff657caf7a3820L92-R94)
* Refactored the `get_json_type` function to return `InputType` values and updated its usage across simulation preparation and testing code. [[1]](diffhunk://#diff-fd6ef04bceceed71a763cfa5341d99a30fd6c65292e0ac6922449b237079b25dL39-R151) [[2]](diffhunk://#diff-fd6ef04bceceed71a763cfa5341d99a30fd6c65292e0ac6922449b237079b25dL70-R171) [[3]](diffhunk://#diff-fd6ef04bceceed71a763cfa5341d99a30fd6c65292e0ac6922449b237079b25dL177-R292) [[4]](diffhunk://#diff-8670b995eb9f58b14f5a7cab50215edce3442835f8cb4d7470ff657caf7a3820L81-R83) [[5]](diffhunk://#diff-8670b995eb9f58b14f5a7cab50215edce3442835f8cb4d7470ff657caf7a3820L92-R94)

**Simulation input file and primaries detection:**
* Added `get_simulator_type_and_particle_source_input_file` and `get_total_number_of_primaries` functions to robustly determine the simulation type, relevant input file, and the total number of primary particles from the payload, improving reliability and maintainability.

**ntasks validation and clamping:**
* Introduced `get_clamped_ntasks_value`, which ensures that the `ntasks` value is always within the valid range `[1, number_of_primaries]`, preventing simulation crashes due to invalid task counts. This function is now called in both batch and celery routes before simulation creation. [[1]](diffhunk://#diff-fbf454bac730996f942faf8a0d5fb46f06ccfaddef50324af9330b13d2256aafR48-R97) [[2]](diffhunk://#diff-fbf454bac730996f942faf8a0d5fb46f06ccfaddef50324af9330b13d2256aafR2-R7) [[3]](diffhunk://#diff-53fdd75abe66e84eaa5bd749cc3688b927d1cd4f36c0ab72461b7a4a75d251deL18-R19) [[4]](diffhunk://#diff-53fdd75abe66e84eaa5bd749cc3688b927d1cd4f36c0ab72461b7a4a75d251deR48-R51) [[5]](diffhunk://#diff-5fd9c29e3abb0af27a7b5c902fe0d58ec296cb83f4db5aedafb496ca2d1b1294L20-R21) [[6]](diffhunk://#diff-5fd9c29e3abb0af27a7b5c902fe0d58ec296cb83f4db5aedafb496ca2d1b1294R49-R52)

**Refactoring of primaries adjustment logic:**
* Refactored `adjust_primaries_in_files_dict`, `adjust_primaries_for_shieldhit_files`, and `adjust_primaries_for_fluka_files` to use the new simulator type and input file detection logic, making these functions more generic and robust to different input file names. [[1]](diffhunk://#diff-fd6ef04bceceed71a763cfa5341d99a30fd6c65292e0ac6922449b237079b25dL106-R224) [[2]](diffhunk://#diff-fd6ef04bceceed71a763cfa5341d99a30fd6c65292e0ac6922449b237079b25dL134-R259) [[3]](diffhunk://#diff-fd6ef04bceceed71a763cfa5341d99a30fd6c65292e0ac6922449b237079b25dL165-R275)

These changes collectively make the simulation submission process safer, more maintainable, and less error-prone.

### DIFF
--- a/yaptide/routes/celery_routes.py
+++ b/yaptide/routes/celery_routes.py
@@ -17,7 +17,7 @@ from yaptide.persistence.models import (CelerySimulationModel, CeleryTaskModel, 
                                         UserModel)
 from yaptide.routes.utils.decorators import requires_auth
 from yaptide.routes.utils.response_templates import (error_validation_response, yaptide_response)
-from yaptide.routes.utils.utils import check_if_job_is_owned_and_exist, determine_input_type, make_input_dict, get_clumped_ntasks_value
+from yaptide.routes.utils.utils import check_if_job_is_owned_and_exist, determine_input_type, make_input_dict, get_clamped_ntasks_value
 from yaptide.routes.utils.tokens import encode_simulation_auth_token
 from yaptide.utils.enums import EntityState, PlatformType
 from yaptide.utils.helper_tasks import terminate_unfinished_tasks
@@ -46,7 +46,9 @@ class JobsDirect(Resource):
             return error_validation_response()
 
         # ensure the ntasks value is in allowed range
-        payload_dict["ntasks"] = get_clumped_ntasks_value(payload_dict=payload_dict, input_type=input_type)
+        payload_dict["ntasks"] = get_clamped_ntasks_value(payload_dict=payload_dict,
+                                                          input_type=input_type,
+                                                          ntasks=payload_dict["ntasks"])
 
         # create a new simulation in the database, not waiting for the job to finish
         job_id = datetime.now().strftime('%Y%m%d-%H%M%S-') + str(uuid4()) + PlatformType.DIRECT.value
--- a/yaptide/routes/celery_routes.py
+++ b/yaptide/routes/celery_routes.py
@@ -40,8 +40,9 @@ class JobsDirect(Resource):
             diff = required_keys.difference(set(payload_dict.keys()))
             return yaptide_response(message=f"Missing keys in JSON payload: {diff}", code=400)
 
-        # ensure there are at most as many tasks as primaries
-        payload_dict["ntasks"] = min(payload_dict["ntasks"], payload_dict["input_json"]["beam"]["numberOfParticles"])
+        # ensure there are at most as many tasks as primaries if using editor
+        if (input_type == "editor" && payload_dict["input_json"] is not None):
+            payload_dict["ntasks"] = min(payload_dict["ntasks"], payload_dict["input_json"]["beam"]["numberOfParticles"])
 
         input_type = determine_input_type(payload_dict)
 
--- a/yaptide/routes/celery_routes.py
+++ b/yaptide/routes/celery_routes.py
@@ -17,7 +17,8 @@ from yaptide.persistence.models import (CelerySimulationModel, CeleryTaskModel, 
                                         UserModel)
 from yaptide.routes.utils.decorators import requires_auth
 from yaptide.routes.utils.response_templates import (error_validation_response, yaptide_response)
-from yaptide.routes.utils.utils import check_if_job_is_owned_and_exist, determine_input_type, make_input_dict, get_clamped_ntasks_value
+from yaptide.routes.utils.utils import (check_if_job_is_owned_and_exist, determine_input_type,
+                                        make_input_dict, get_clamped_ntasks_value)
 from yaptide.routes.utils.tokens import encode_simulation_auth_token
 from yaptide.utils.enums import EntityState, PlatformType
 from yaptide.utils.helper_tasks import terminate_unfinished_tasks
--- a/yaptide/routes/celery_routes.py
+++ b/yaptide/routes/celery_routes.py
@@ -42,11 +42,11 @@ class JobsDirect(Resource):
 
         input_type = determine_input_type(payload_dict)
 
-        # ensure the ntasks value is in allowed range
-        payload_dict["ntasks"] = get_clumped_ntasks_value(payload_dict=payload_dict, input_type=input_type)
-
         if input_type is None:
             return error_validation_response()
+
+        # ensure the ntasks value is in allowed range
+        payload_dict["ntasks"] = get_clumped_ntasks_value(payload_dict=payload_dict, input_type=input_type)
 
         # create a new simulation in the database, not waiting for the job to finish
         job_id = datetime.now().strftime('%Y%m%d-%H%M%S-') + str(uuid4()) + PlatformType.DIRECT.value
--- a/yaptide/routes/celery_routes.py
+++ b/yaptide/routes/celery_routes.py
@@ -41,7 +41,7 @@ class JobsDirect(Resource):
             return yaptide_response(message=f"Missing keys in JSON payload: {diff}", code=400)
 
         # ensure there are at most as many tasks as primaries if using editor
-        if (input_type == "editor" && payload_dict["input_json"] is not None):
+        if (payload_dict["input_type"] == "editor" and payload_dict["input_json"] is not None):
             payload_dict["ntasks"] = min(payload_dict["ntasks"], payload_dict["input_json"]["beam"]["numberOfParticles"])
 
         input_type = determine_input_type(payload_dict)
--- a/yaptide/routes/celery_routes.py
+++ b/yaptide/routes/celery_routes.py
@@ -17,7 +17,7 @@ from yaptide.persistence.models import (CelerySimulationModel, CeleryTaskModel, 
                                         UserModel)
 from yaptide.routes.utils.decorators import requires_auth
 from yaptide.routes.utils.response_templates import (error_validation_response, yaptide_response)
-from yaptide.routes.utils.utils import check_if_job_is_owned_and_exist, determine_input_type, make_input_dict
+from yaptide.routes.utils.utils import check_if_job_is_owned_and_exist, determine_input_type, make_input_dict, get_clumped_ntasks_value
 from yaptide.routes.utils.tokens import encode_simulation_auth_token
 from yaptide.utils.enums import EntityState, PlatformType
 from yaptide.utils.helper_tasks import terminate_unfinished_tasks
@@ -40,11 +40,10 @@ class JobsDirect(Resource):
             diff = required_keys.difference(set(payload_dict.keys()))
             return yaptide_response(message=f"Missing keys in JSON payload: {diff}", code=400)
 
-        # ensure there are at most as many tasks as primaries if using editor
-        if (payload_dict["input_type"] == "editor" and payload_dict["input_json"] is not None):
-            payload_dict["ntasks"] = min(payload_dict["ntasks"], payload_dict["input_json"]["beam"]["numberOfParticles"])
-
         input_type = determine_input_type(payload_dict)
+
+        # ensure the ntasks value is in allowed range
+        payload_dict["ntasks"] = get_clumped_ntasks_value(payload_dict=payload_dict, input_type=input_type)
 
         if input_type is None:
             return error_validation_response()
--- a/yaptide/routes/celery_routes.py
+++ b/yaptide/routes/celery_routes.py
@@ -40,6 +40,9 @@ class JobsDirect(Resource):
             diff = required_keys.difference(set(payload_dict.keys()))
             return yaptide_response(message=f"Missing keys in JSON payload: {diff}", code=400)
 
+        # ensure there are at most as many tasks as primaries
+        payload_dict["ntasks"] = min(payload_dict["ntasks"], payload_dict["input_json"]["beam"]["numberOfParticles"])
+
         input_type = determine_input_type(payload_dict)
 
         if input_type is None:
--- a/yaptide/routes/celery_routes.py
+++ b/yaptide/routes/celery_routes.py
@@ -47,7 +47,6 @@ class JobsDirect(Resource):
 
         # ensure the ntasks value is in allowed range
         payload_dict["ntasks"] = get_clamped_ntasks_value(payload_dict=payload_dict,
-                                                          input_type=input_type,
                                                           ntasks=payload_dict["ntasks"])
 
         # create a new simulation in the database, not waiting for the job to finish
--- a/yaptide/routes/utils/utils.py
+++ b/yaptide/routes/utils/utils.py
@@ -3,7 +3,7 @@ from typing import Optional
 from yaptide.persistence.db_methods import fetch_simulation_by_job_id
 from yaptide.persistence.models import UserModel
 from yaptide.utils.enums import InputType, SimulationType
-from yaptide.utils.sim_utils import files_dict_with_adjusted_primaries, get_primaries_file_type_and_name
+from yaptide.utils.sim_utils import files_dict_with_adjusted_primaries, get_total_number_of_primaries
 
 
 def check_if_job_is_owned_and_exist(job_id: str, user: UserModel) -> tuple[bool, str, int]:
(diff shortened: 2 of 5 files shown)


Mosder:
Just realized it's not as easy of a fix for simulation using files instead of an editor. I don't actually know if it's even possible without having to delete files (I don't know exactly how the simulation is being represented by the files) which is a no-go. I'll try to look into it today evening.

yaptide/routes/celery_routes.py
```diff
@@ -40,6 +40,10 @@ def post(user: UserModel):
             diff = required_keys.difference(set(payload_dict.keys()))
             return yaptide_response(message=f"Missing keys in JSON payload: {diff}", code=400)
 
+        # ensure there are at most as many tasks as primaries if using editor
+        if (payload_dict["input_type"] == "editor" and payload_dict["input_json"] is not None):
+            payload_dict["ntasks"] = min(payload_dict["ntasks"], payload_dict["input_json"]["beam"]["numberOfParticles"])
+
```
grzanka: The logic is hard to understand for human when written this way, I'd prefer to be more explicit, explaining in the comment **the reason** for particular behaviour. Here the non-obvious reason is that simulations crash for zero primaries.

```
# in case user requested less primaries that tasks, we need to avoid submission of tasks with zero primaries, as this may crash 
# for example: 7 primaries on 11 tasks requested means that there will be 4 tasks with zero primaries (which will crash)
# to avoid crash we reduce number of tasks, so we zero-primaries tasks are excluded and rests of tasks handle one primarie
# in the example of 7 primaries on 11 tasks, we reduce number of tasks from 11 to 7

if (payload_dict["input_type"] == "editor" and payload_dict["input_json"] is not None):
    if payload_dict["ntasks"] > payload_dict["input_json"]["beam"]["numberOfParticles"]:  
          payload_dict["ntasks"] = payload_dict["input_json"]["beam"]["numberOfParticles"]
```

grzanka (changes requested):
See comment inline

Mosder:
I added the validation for the simulation using files and moved it to a separate function in utils which clearly explains what it does. I also cleaned up a code to be more readable.

grzanka:
Below a Linus style AI-generated review (I fully agree with it):

*cracks knuckles*

---

Christ on a bike. Who wrote this steaming pile and thought it was acceptable to push?

Let me walk you through this disaster, because apparently the author couldn't be bothered to think for more than five seconds before hitting commit.

**1. The function name is already wrong.**

"Clumped"? The word you're looking for is **"clamped"**, you absolute walnut. `clamp` is the standard term used in every math library since the dawn of computing. It's like naming your function `get_scrunched_value` instead of `get_compressed_value`. Did you even proofread this, or did you just sneeze on the keyboard and ship it?

**2. You call `next()` with the same generator expression TWICE.**

```python
elif next((file for file in payload_dict["input_files"] if file.endswith(".inp")), None):
    input_file = next((file for file in payload_dict["input_files"] if file.endswith(".inp")), None)
```

You iterate the entire dictionary TWICE to find the same damn file. What, was a local variable too fancy for you? This isn't just ugly — it's the kind of lazy thinking that tells me you copy-pasted from the `elif` into the body and called it a day. How about:

```python
input_file = next((file for file in payload_dict["input_files"] if file.endswith(".inp")), None)
elif input_file is not None:
```

Was that so bloody hard?

**3. Zero defensive programming. None of it.**

- What happens when `all_beam_lines_with_nstat` is **empty**? You blindly do `[0].split()[1]`. Boom. `IndexError`. Hope you enjoy your 3 AM pager alert.
- What happens when `start_card` is `None`? You call `.split()[1]` on `None`. Boom. `AttributeError`. Brilliant.
- What happens when `number_of_all_primaries` is **zero** or **negative** from some garbage input file? You'll happily return 0 or -3 as a "valid" number of tasks. So much for that beautiful docstring explaining why ntasks must be ≥ 1.

You wrote a whole paragraph in the docstring about why values must be in `[1, primaries_count]` and then **didn't actually enforce the lower bound on primaries_count**. The docstring is literally a better programmer than you.

**4. The function does way too much.**

This thing is parsing three different file formats, validating a value, AND clamping it. Ever heard of the Single Responsibility Principle, or did you skip that day? The primaries extraction should be its own function (or three — one per format). This function should just clamp a value. That's it.

**5. Raw `dict` access everywhere with zero error handling.**

`payload_dict["ntasks"]`, `payload_dict["input_json"]["beam"]["numberOfParticles"]`, `payload_dict["input_files"]['beam.dat']`... Any one of these can throw a `KeyError` and you've done absolutely nothing about it. You're treating an external input dictionary like it's a blessed, immutable contract from God. It's not. It's user input. It **will** be wrong.

**6. `int(float(start_card.split()[1]))`**

This is the kind of thing that makes me mass-reply NACK to pull requests. If the value is supposed to be an integer, why is it a float in the file? If there's a reason, **put it in a comment**. If there isn't, don't do the double conversion. This screams "I don't know why this works but it does so I'm not touching it."

---

In summary: rename it to `clamp`, extract the parsing logic, add error handling that isn't "pray and hope," and stop iterating the same collection twice like some kind of O(2n) enthusiast. The docstring promises a contract this code doesn't even remotely deliver on.

I'm not merging this. Fix it and come back when it doesn't make me want to mass-revert the entire branch.

grzanka (changes requested):
See comments below the code

Mosder:
Ok, here's what I have to say to you Linus:
1. Ah yes, a typographical mistake, something worth getting it's own bullet point and something worth losing your mind over. Yes, it's my bad, but if you're gonna overreact like that for every typo you ever see, that's all you gonna do. You need to pump the brakes a bit, lest your hair turns gray... Oh wait, it already did.
2. Yes, I did use next twice. Yes, it's ugly. But since when do you care about such things when the reason that's what it is, is because I copied the way it was already being done in the codebase. Surely you wouldn't have overlooked such a thing, so you must have green-lit it, right?
3. That is true, I didn't add any defensive programming, but what's the purpose? What's the purpose of checking any of that when other functions in the codebase don't. Sure, they check some things, but not all, and not checking all is like checking nothing.
4. Oh, I certainly did hear of it. The thing is, one look at the codebase suggests that's not how we do it here. I mean, some of the functions I've seen basically build a house, plant a tree and raise a son by themselves. Like, congratulations to them for being a textbook man, but c'mon, leave something to do for the rest of us.
5. Again, like half of the functions in **yaptide/utils/sim_utils.py** do some unsafe operations, so adding safe access here is not gonna change the fact that we're still just praying for no errors.
6. THAT'S WHAT I'D LIKE TO KNOW. It's not like I came up with some random way of parsing the value, it's LITERALLY how it's done in the `adjust_primaries_for_fluka_files` function and wouldn't you guess it, there's no comment to explain the reasoning behind it. So, your guess is as good as mine here.

In conclusion, I could say the same thing to you that you did to a certain company with a green logo, but I'm not going to, since I respect the elderly. All I'm gonna say is: you can't pour from an empty cup (man, the Polish version of this idiom sounds way better).

PS:  Since when are you a python expert? Don't you literally vibe code python, so you don't have to write in it? Which I can respect by the way, but it kinda undermines your authority here, y'know?

PPS: C'mon Linus O(2n)? You know better than that... Or did you already go senile and are forgetting how the big O notation works. Saying O(2n) is like saying "most optimal". We know what you mean, but please, use proper terminology, lest you make a fool of yourself.

yaptide/routes/utils/utils.py
```diff
@@ -44,3 +44,44 @@ def make_input_dict(payload_dict: dict, input_type: str) -> dict:
     input_dict["input_files"] = files_dict
 
     return input_dict
+
+
+def get_clamped_ntasks_value(payload_dict: dict, ntasks: int) -> int:
+    """
+    Function that validates ntasks value in a simulation and returns the number of tasks that the simulation should use.
+    It's used to prevent simulations with ntasks outside of range [1; primaries_count].
+    The ntasks value needs to be larger than 0, because you need at least a single task to run a simulation.
```
grzanka: To be exact: 
```suggestion
    The ntasks value needs to be larger than 0, because you need at least a single task to run a simulation with certain simulators (like SHIELD-HIT12A).
```

yaptide/routes/utils/utils.py
```diff
@@ -44,3 +44,44 @@ def make_input_dict(payload_dict: dict, input_type: str) -> dict:
...
+
+    Args:
+        payload_dict: A dictionary containing the payload received from a request.
+        ntasks: Task count from the request to be clamped
+
+    Returns:
+        Integer value representing the ntasks value to use in the simulation.
+    """
+    # ensure there is at least 1 task
+    if ntasks < 1:
```
grzanka, line 68: at that moment it would be worth to add some warning message, to be printed in the logs, that user did something stupid, requesting 0 or negative number of tasks

yaptide/routes/utils/utils.py
```diff
@@ -44,3 +44,44 @@ def make_input_dict(payload_dict: dict, input_type: str) -> dict:
...
+        return 1
+
+    # get the number of primaries
```
grzanka:
```suggestion
    # get the total number of primaries for whole simulation/job
```

yaptide/routes/utils/utils.py
```diff
@@ -44,3 +44,44 @@ def make_input_dict(payload_dict: dict, input_type: str) -> dict:
...
+    number_of_all_primaries = get_total_number_of_primaries(payload_dict)
+
+    # if we couldn't get the total number of primaries, fall back to the original ntasks value
+    if number_of_all_primaries is None:
```
grzanka, line 76: again, a warning here would help later with debugging

yaptide/utils/sim_utils.py
```diff
@@ -36,18 +37,111 @@ def estimators_to_list(estimators_dict: dict, dir_path: Path) -> list[dict]:
...
+def get_json_type(payload_dict: dict) -> InputType:
+    """Returns type of provided JSON"""
+    if "input_files" in payload_dict:
+        return InputType.FILES
+    return InputType.EDITOR
 
-    Editor = auto()
-    Files = auto()
 
+def get_primaries_file_type_and_name(payload_dict: dict) -> tuple[Optional[SimulationType], Optional[str]]:
```
grzanka: We rather call that file (or file fragment) "particle source" than "primaries", so I'd rather call this `get_simulator_type_and_particle_source_input_file`

yaptide/utils/sim_utils.py
```diff
@@ -36,18 +37,111 @@ def estimators_to_list(estimators_dict: dict, dir_path: Path) -> list[dict]:
...
+    Args:
+        payload_dict: A dictionary containing the payload received from a request.
+
+    Returns:
+        Tuple of SimulationType enumerator and string containing the name of the file
+        or the tuple of Nones if the file type couldn't be determined
+    """
+    # ensure that the payload contains the input files dictionary
+    input_files = payload_dict.get("input_files")
+    if not input_files:
```
grzanka, line 62: warning here would be useful

yaptide/utils/sim_utils.py
```diff
@@ -36,18 +37,111 @@ def estimators_to_list(estimators_dict: dict, dir_path: Path) -> list[dict]:
...
+
+    # determining input file type - that information should most certainly be passed in the payload,
+    # but now it's not, so we have to do this ugly thing
+    if "beam.dat" in input_files:
+        return SimulationType.SHIELDHIT, "beam.dat"
+    inp_file = next((file for file in input_files if file.endswith(".inp")), None)
+    if inp_file:
+        return SimulationType.FLUKA, inp_file
+
+    # if we couldn't determine the file type, return a tuple of Nones
```
grzanka, line 74: warning here, please

yaptide/utils/sim_utils.py
```diff
@@ -36,18 +37,111 @@ def estimators_to_list(estimators_dict: dict, dir_path: Path) -> list[dict]:
...
+    """
+    input_type = get_json_type(payload_dict)
+
+    # get number of primaries when EDITOR was used
+    if input_type == InputType.EDITOR:
+        # check with try-catch for the number of particles variable and return it
+        # if it's not there, return None
+        try:
+            return payload_dict["input_json"]["beam"]["numberOfParticles"]
+        except KeyError:
```
grzanka, line 98: a warning here please

yaptide/utils/sim_utils.py
```diff
@@ -103,20 +197,29 @@ def adjust_primaries_in_files_dict(payload_files_dict: dict, ntasks: int = None)
...
+        return adjust_primaries_for_fluka_files(payload_files_dict=payload_files_dict,
+                                                ntasks=ntasks,
+                                                primaries_file=file_name)
     return {}, 0
 
 
-def adjust_primaries_for_shieldhit_files(payload_files_dict: dict, ntasks: int = None) -> tuple[dict, int]:
+def adjust_primaries_for_shieldhit_files(payload_files_dict: dict,
+                                         ntasks: int = None,
+                                         primaries_file: str = None) -> tuple[dict, int]:
```
grzanka: I'd rather call than `beam_dat_file` -> its SHIELD-HIT12A specific name

grzanka:
```suggestion
                                         primaries_file: str = "beam.dat") -> tuple[dict, int]:
```

why not using that as default, instead of `if not primaries_file`... logic ?

yaptide/utils/sim_utils.py
```diff
@@ -131,22 +234,27 @@ def adjust_primaries_for_shieldhit_files(payload_files_dict: dict, ntasks: int =
...
     # number_of_tasks = payload_files_dict['ntasks']  -> to be implemented in UI
     # here we manipulate the files_dict['beam.dat'] file to adjust number of primaries
     # we manipulate content of the file, no need to write the file to disk
     return files_dict, int(number_of_all_primaries)
 
 
-def adjust_primaries_for_fluka_files(payload_files_dict: dict, ntasks: int = None) -> tuple[dict, int]:
+def adjust_primaries_for_fluka_files(payload_files_dict: dict,
+                                     ntasks: int = None,
+                                     primaries_file: str = None) -> tuple[dict, int]:
```
grzanka: Here I'd use `fluka_input_file`

grzanka (changes requested):
Very nice overall, few comments inline

grzanka:
@Mosder code looks good, now I need to test it on yap-dev.

grzanka:
https://github.com/user-attachments/assets/e98a2ff9-4c4e-43a8-8e24-91710dc2467b

@Mosder 
Is it intentional that simulation with 9 primaries and 10 tasks fails ?

I've deployed this PR to yap-dev, you can test it at https://yap-dev.c3.plgrid.pl/

Mosder:
That's weird, it definitely should work. It works for me locally. I'll look into this tomorrow.

grzanka:
> That's weird, it definitely should work. It works for me locally. I'll look into this tomorrow.

Lets discuss then on Zulip:

image

grzanka (approved):

grzanka:
OK, my bad, I had some troubles in deployment

PR #1635, opened 2026-03-24
Add ELK stack

lythx:
This PR adds a `docker-compose.elk.yml` file which can be used in docker setup to enable the Elasticsearch + Logstash + Kibana logging stack. To test it run:

`docker compose -f docker-compose.yml -f docker-compose.elk.yml -f docker-compose.fast.yml up`

I used `gelf` driver to copy the stdout of yaptide processes to Elasticsearch. With this approach we don't need to modify the source code (the only change is the new `docker-compose.elk.yml` file), however this splits logs line-by-line which is inconvenient for multi-line logs. I tried to fix it by changing gelf config, but it didn't work (mostly due to the fact that the flask process has different log format than the workers, and startup logs don't follow the common log format). According to AI the proper way to solve this is to use `python-json-logger` library, but that requires rewriting some logging code (it does have considerable benefits though, e.g. easy filtering by simulation id/task id/app instance).


grzanka:
@lythx https://github.com/yaptide/yaptide/pull/1642 this will affect your work

PR #1637, opened 2026-03-25
Fix failed simulation handling

### LINKED ISSUE #1636: test_run_simulation_with_flask_crashing fails sometimes
Locally it fails around 1/50 times, on GHA more often (like 1/3 I think).

lythx:
This PR implements better handling of failed simulations, specifically the case in which ALL tasks fail. Currently, if that happens, the job gets stuck in `MERGING_RUNNING` state. I added a `task_state` key to object passed between simulation tasks. `merge_results` task marks the job as `FAILED` if every single result has `task_state != COMPLETED`. I also changed the order of operations in `merge_results` so now the logfiles get sent to the backend even in case of complete failure.


yaptide/celery/tasks.py
```diff
@@ -231,14 +237,46 @@ def set_merging_queued_state(results: list[dict]) -> list[dict]:
...
+        logging.info("Received empty or None 'results' list in merge_results(): %s", results)
+        return final_result
+
+    # find simulation_id and update_key for sending updates to the backend and get all logfiles from the results
+    simulation_id = None
+    update_key = None
     logfiles = {}
+    for i, result in enumerate(results):
+        if simulation_id is None:
+            simulation_id = result.pop("simulation_id", None)
```
grzanka: why `pop` and not `get` ?

grzanka (changes requested):

grzanka:
@lythx itseems there is some copilot review stuff left

grzanka:
@lythx https://github.com/yaptide/yaptide/pull/1642 this will affect your work

grzanka:
@lythx fix the conflicts

Issue #413, opened 2023-05-24
Reconfigure celery to support multiple instances


grzanka:
Celery uses single worker now, which sits in a single docker container.
See here:
https://github.com/yaptide/yaptide/blob/master/yaptide/celery/worker.py

We should reconfigure backend to use pool of many workers, either created manually or in kuberentes (as multiple pods)

grzanka:
See also https://github.com/yaptide/yaptide/issues/629

grzanka:
We would like to use multiple instances in C3 cloud

PR #1648, opened 2026-05-12
Add support for using multiple instances with Celery

Mosder:
This PR aims to add support for using multiple instances with Celery (issue #413). It modifies the current docker-compose.yml file, so that it exposes Redis and Flask to the local network and adds a password for using Redis. Creates a docker-compose-simulation_worker.yml which can be used to activate simulation-worker on a separate instance. For the deployment to work, both the main host machine and all additional instances need to define the HOST_IP (local ip of the main host machine) and REDIS_PASSWORD (password for using Redis) environment variables.


grzanka:
@Mosder can you merge main/master branch so the tests could run ?

grzanka:
Thanks @Mosder for taking this on — #413 has been open for a long time. I went through the change carefully and also verified a few of the behaviours below against real containers rather than guessing.

The deployment story splits cleanly into two cases, so let me structure the review that way, and then list the issues that apply in both.

---

## Case A — the two instances **cannot** see each other over the network

**This does not work, and cannot be made to work with the current design.**

The worker is a pure *outbound-connect* client in three directions, all toward the main host:

- `CELERY_BROKER_URL=redis://…@${HOST_IP}:6379/0` — worker dials the main host
- `CELERY_RESULT_BACKEND=…` — same
- `BACKEND_INTERNAL_URL=http://${HOST_IP}:6000` — worker POSTs status / results / logfiles to Flask (`yaptide/celery/utils/requests.py`)

So the worker's network must be able to route to the main host on 6379 and 6000. Nothing in the design reverses that — there is no reverse tunnel, no worker-side listener, no relay. If the two sites can't reach each other, no value of `HOST_IP` helps.

There is also a design smell in the variable itself: `HOST_IP` is simultaneously the **bind address** on the main host (`ports: "${HOST_IP}:6379:6379"`) and the **dial address** on the worker. Those are the same string only on a flat LAN. Behind NAT the worker needs a public/VPN address while the main host must bind a private one — one knob is being asked to play two roles, so anything with address translation breaks it.

Options, roughly in order of how well they fit this repo:

1. **Overlay network (WireGuard / Tailscale / ZeroTier).** Both hosts get a stable overlay address, `HOST_IP` becomes the main host's overlay IP, and the traffic is encrypted for free — which also resolves Case B below. This is by far the smallest change and I'd suggest making it *the* documented deployment mode for this feature.
2. **SSH reverse tunnel** from the worker site (`ssh -R`), if only one side can initiate connections.
3. Worth noting the precedent already in the codebase: yaptide already runs remote compute across unroutable networks via SSH batch submission (`yaptide/batch/batch_methods.py`, the `Clusters` route). That works precisely because it is outbound SSH from the backend. "Celery over an open LAN port" is a materially different trust posture and should be an explicit, documented decision rather than an implicit one.

---

## Case B — the two instances **can** see each other

It works, but **none of the traffic is encrypted**:

- `redis://` is plaintext TCP. `rediss://` would be TLS, but the redis service is not given `--tls-port` / certificates, and the Celery app (`yaptide/celery/simulation_worker.py`) sets no `broker_use_ssl` / `redis_backend_use_ssl`.
- The Redis `AUTH` password added by this PR is itself sent **in cleartext** on every connection. It protects against a random port scanner, not against anyone who can sniff the segment.
- Celery task payloads carry the simulation input files **and the `update_key` JWT** — in the clear.
- Results and logfiles are POSTed to `http://${HOST_IP}:6000` — plaintext HTTP, and this **bypasses the nginx TLS terminator** (`nginx.conf` listens on 5000/8443 and proxies to `yaptide_flask:6000`). Publishing 6000 exposes the *entire* API unencrypted, `/auth/login` included, so user credentials and JWTs become sniffable as well.

Two further points specific to this case:

**B1 — Insecure-by-default: a plain `docker compose up` now opens Redis to the world.** Both new variables fail open, and I checked both:

```
$ docker run redis:8-alpine redis-server --requirepass ""   →  redis-cli ping → PONG     (no auth at all)
$ HOST_IP= docker compose config                            →  published: "6379"        (i.e. 0.0.0.0)
```

So anyone who pulls master after this merges and does not set the new variables gets an **unauthenticated Redis and the raw Flask API published on every interface**, where before this PR neither port was published at all. That is a silent regression for every existing deployment. Suggested fix: `${REDIS_PASSWORD:?REDIS_PASSWORD must be set}` and `${HOST_IP:-127.0.0.1}`, so it either refuses to start or stays loopback-only.

**B2 — A reachable Redis is an RCE surface, not merely a data-leak surface.** Anyone who reaches 6379 (or sniffs the password) can inject arbitrary messages into the `simulations` queue → arbitrary code execution in the workers; read every task payload; harvest `update_key` JWTs and forge results via `/tasks`, `/results`, `/logfiles` (those endpoints authenticate on `update_key` alone — `yaptide/routes/task_routes.py`, `yaptide/routes/common_sim_routes.py`); `FLUSHALL`; or `CONFIG SET dir/dbfilename` to write files. The PR uses the `default` user with the full command set — a dedicated ACL user with a restricted command list would be cheap hardening.

**How to actually encrypt it**, in increasing order of effort: overlay VPN (transparent, no code change) → stunnel → real Redis TLS (`--tls-port 6379 --port 0` plus certificates, `rediss://` URLs, `broker_use_ssl` / `redis_backend_use_ssl` on the Celery app) together with pointing `BACKEND_INTERNAL_URL` at nginx's `https://…:8443`. That last step needs a CA the worker trusts, so self-signed certificates will need handling.

---

## Issues that apply in both cases

**1 — All workers share one hardcoded Celery node name.** `run_simulation_worker.sh` pins `--hostname yaptide-simulation-worker`; I confirmed this resolves to `celery@yaptide-simulation-worker` for every container on every host. Consequences: (a) the `celery … inspect ping` healthcheck broadcasts and gets a reply from *any* worker in the fleet, so a hung local worker still reports healthy and `restart: unless-stopped` never fires; (b) mingle/gossip/events and control-reply routing are confused by duplicate node names; (c) monitoring cannot tell nodes apart. Should be `--hostname simulation-worker@%h`. Since #413 is precisely about running a pool of workers, this is arguably the one code change the PR most needs, and it isn't included.

**2 — The new compose file drops the `simulators` volume.** `docker-compose.yml` mounts `simulators:/simulators:rw`; `docker-compose-simulation_worker.yml` has no volumes. Because `run_simulation_worker.sh` downloads SHIELD-HIT / TOPAS / FLUKA at container start, every restart re-downloads them (TOPAS is multi-GB from S3) into the container layer. Combined with `restart: unless-stopped`, a crash-looping worker becomes a download loop.

**3 — Duplicated service definition instead of an override.** The worker service is now defined in two places and the copies have already diverged (see point 2). A `docker compose -f docker-compose.yml -f docker-compose-remote-worker.yml` override, or `profiles:`, would keep a single source of truth. Related: `container_name: yaptide_simulation_worker` is fixed, so more than one worker container per remote host is impossible (`--scale` fails) — a limitation worth stating explicitly in a "multiple instances" PR.

**4 — `--appendonly yes` on a service with no volume.** The redis service has no volume mount, so the AOF is written into the container's writable layer and lost on every recreate, while growing that layer in the meantime. Either add a named volume or drop the flag.

**5 — Secret distribution isn't addressed.** `env_file: .env` on the remote worker means the *entire* main `.env` (S3 keys, the SHIELD-HIT decryption key, the Postgres password) has to be copied to every worker machine, and `REDIS_PASSWORD` baked into `CELERY_BROKER_URL` is visible to anything that can run `docker inspect`. A worker-only env file would be better.

**6 — Dead configuration.** `environment: REDIS_PASSWORD` / `HOST_IP` on the redis service and `HOST_IP` on `yaptide_flask` are unused — the redis image doesn't read `REDIS_PASSWORD` (the password comes from `--requirepass` in `command:`), and Flask never reads `HOST_IP`. Worth removing so they don't mislead the next reader.

**7 — No documentation, no CI coverage.** Nothing outside the compose files mentions `HOST_IP` or `REDIS_PASSWORD` — no README section, no `.env.example`, no note that 6379/6000 must be opened in the firewall. No workflow runs `docker compose`, so nothing exercises this path at all, and the PR description is currently the only documentation.

---

## Summary

The change does what the description says **on a flat, trusted LAN with the firewall opened**, and it's a reasonable first step toward #413. Even in that setting I'd consider points B1 (fail-open defaults) and 1 (duplicate node names defeating the healthcheck) blocking. For the "instances cannot see each other" case it does not help at all, and in the "they can see each other" case everything — task payloads, JWTs, the new Redis password, and now the whole HTTP API including login — crosses the wire in plaintext.

Concrete suggestion for the next iteration, keeping the scope as-is:

- make the new variables required rather than fail-open,
- fix `--hostname` to use `%h`,
- restore the simulators volume,
- document the deployment as *"workers must share a private/VPN network with the backend"*, with WireGuard or Tailscale as the recommended route — that single choice makes it both routable across sites and encrypted, and costs far less than bolting TLS onto Redis and nginx.

PR #1661, opened 2026-08-04
Local slurm

lythx:
This PR adds a local slurm setup. Don't worry about the ca_key keyfiles being in the repo - it's needed for local mock certificate authority to work, not a security threat - used only locally.

I copied [this](https://github.com/giovtorres/slurm-docker-cluster) slurm setup (MIT license), and added Keycloak and Certificate Authority services to docker-compose.slurm.yml. The setup emulates PLGrid infrastructure with minimal codebase changes:
- I added "port" column to ClusterModel (local slurm ssh runs on port 3022 by default to avoid collision port collisions)
- I made flask pass BACKEND_EXTERNAL_URL to celery worker responsible for communicating with slurm. It's used only when environment variable is missing (which is the case in the local setup) so it wont't impact dockerized production setups

Setup instructions are in the [docs](https://yaptide.github.io/for_developers/local-setup/local-slurm/)

Yaptide backend must be running locally on host 0.0.0.0 (in order to let slurm reach it). Dockerized yaptide + slurm is not supported.

The setup creates a Keycloak instance and a mock certificate authority server. The mock certificate authority uses the private key `/slurm/ca_key/ca_key` to sign the certs. Entrypoint script puts the public key `/slurm/ca_key/ca_key.pub` into the Slurm cluster and configures it to trust any certificates signed by that authority.

The docker compose and setup script do the following things:
- Set up auth 
- Mount shared `/tmp/scratch` on all nodes
- Add **$SCRATCH** variable `/tmp/scratch/{USER}`
- Create devuser in the Slurm cluster
- Add the cluster to yaptide database
- Copy the Shieldhit binary into the cluster
- Create a **LMOD** module for Shieldhit
- Install **pymchelper** library on all nodes (required for merging results)


grzanka:
@lythx can you merge main/master branch so the tests could run ?

slurm/readme/README_CN.md
```diff
@@ -0,0 +1,168 @@
+# Slurm Docker Cluster
+
+<p align="center">
+    <b> <a href="../README.md">English</a> | 简体中文 </b>
+</p>
+
+**Slurm Docker Cluster**借助Docker Compose实现多容器Slurm集群的敏捷部署。本仓库简化了
```
grzanka, line 7: why in chinese ?

slurm/examples/jobs/gpu_test.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+#SBATCH --job-name=gpu-test
+#SBATCH --partition=gpu
+#SBATCH --gres=gpu:1
+#SBATCH --output=gpu-test-%j.out
```
grzanka, line 5: what is that ? we don't need GPU ....

migrations/versions/d0a89a16fbac_add_port_column_to_cluster.py
```diff
@@ -0,0 +1,31 @@
...
+revision = 'd0a89a16fbac'
+down_revision = '5003b9acb1f4'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('Cluster', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('port', sa.Integer(), nullable=False))
```
grzanka, line 22: why do we need this port stuff ?

grzanka (commented):
see comments

grzanka (commented):
## Critical review

Went through the diff in detail (migrations, `slurm/*`, `yaptide/batch/batch_methods.py`, compose/Makefile/scripts). Two issues look serious enough to block merge; the rest are worth fixing but not blockers.

### 🔴 Blocking

**1. Migration will fail on any DB that already has `Cluster` rows** — `migrations/versions/d0a89a16fbac_add_port_column_to_cluster.py:22`
The new `port` column is added `nullable=False` with no `server_default`. `ALTER TABLE Cluster ADD COLUMN port INTEGER NOT NULL` fails on SQLite ("Cannot add a NOT NULL column with default value NULL") and on Postgres ("column contains null values") as soon as one cluster row already exists — i.e. on every non-fresh deployment. The ORM-level `default=22` in `yaptide/persistence/models.py` doesn't help here since Alembic DDL only respects `server_default`. Earlier migrations in this repo (`5003b9acb1f4_.py`, `41149fa64bd2_.py`) pair `nullable=False` with `server_default=...` for exactly this reason.

**2. Mock CA never verifies the JWT signature** — `slurm/certificate_authority.py:28`
```python
jwt.decode(token, options={"verify_signature": False})
```
followed by trusting the `preferred_username` claim to pick the SSH cert principal. Anyone who can reach the CA's published port (5001, `docker-compose.slurm.yml`) can send `Authorization: Bearer <unsigned forged JWT with preferred_username=root>` and get back a CA-signed SSH certificate for `root`, trusted by `slurmctld`'s sshd via `TrustedUserCAKeys`. Even for a "mock"/dev CA this is worth gating behind an explicit dev-only flag or at least a comment, since it's the kind of thing that quietly ships into a shared environment.

Related, same file (line 30): the JWT decode is wrapped in a bare `except Exception` that falls through to minting a certificate for a hardcoded fallback user `devuser` instead of rejecting the request — a malformed/garbage bearer token gets a valid signed cert rather than a 400/401.

### 🟠 Worth fixing

**3. `BACKEND_EXTERNAL_URL` isn't actually wired to where it's read** — `yaptide/batch/batch_methods.py:232`
`prepare_script_files` reads `BACKEND_EXTERNAL_URL` from `os.environ`, but `docker-compose` only maps that var into the `yaptide_flask` service, not into the celery worker container that actually executes this code — so it resolves to `""` on the real deployment path this PR is meant to support. Notably, an earlier commit on this branch (`0e2d33d`) threaded a `backend_external_url` parameter through `submit_job`/`prepare_script_files` to fix exactly this, but the final commit (`4ccb276`, "remove backend_external_url parameter") reverted it. Worth double-checking that revert was intentional — `watcher.py` / `simulation_data_sender.py` silently no-op (`if not backend_url: return False`) when it's empty, so job status/result delivery would quietly stop working with no visible error.

Also line 234: leftover `logging.warning(backend_url)` fires at WARNING severity on *every* job submission — looks like debug output from tracking down the above, not intentional logging.

**4. `get_connection` now always passes `port=cluster.port` to Fabric** — `yaptide/batch/batch_methods.py:64`
If any existing `cluster_name` embeds a `user@host:port` shorthand (Fabric's `derive_shorthand` supports this, and it was the only way to set a non-default SSH port before this PR), `Connection.__init__` raises `ValueError: You supplied the port via both shorthand and kwarg!` once both are present. Might be worth a quick check for existing cluster rows using that shorthand before this ships.

**5. Fail-fast startup check was dropped** — `slurm/docker-entrypoint.sh` (~line 107)
The old hard check that aborted container startup when the CA public key or `authorized_keys` were missing was replaced by two independent `if` blocks with no `else`/failure path. If the `ca_key.pub` bind mount is missing, `slurmctld` now starts "successfully" with SSH auth silently non-functional, and the failure only surfaces later as an opaque Fabric/paramiko error during job submission.

**6. Replica-number detection has a silent collision fallback** — `slurm/docker-entrypoint.sh` (~line 34)
`detect_replica_number()` falls back to node name `"1"` if its DNS-guessing loop fails to resolve the container's own IP. Under a DNS-propagation race at boot, or once scaled past the hardcoded `max_replicas=64` (the Makefile explicitly documents `make scale-cpu-workers N=100`), two workers could register under the identical `NodeName=c1`.

### 🟡 Minor / polish

- `slurm/.env.example`: `SLURM_SSH_AUTHORIZED_KEYS` is documented but nothing in `docker-compose.slurm.yml` mounts it to `/tmp/authorized_keys_host`, so setting it currently does nothing.
- `scripts/setup_local_slurm.sh:4`: hardcodes `-p 3022` for `add-cluster` instead of reading `SLURM_SSH_PORT`, so overriding that env var breaks the DB's stored port silently.
- `slurm/Makefile`: `set-version` (and `build-all`) overwrite `.env` with `>` instead of updating just the `SLURM_VERSION` line, wiping out any other custom `.env` settings (`SLURM_SSH_PORT`, `SLURM_SSH_AUTHORIZED_KEYS`, etc.).
- `slurm/Makefile`: `.PHONY`/`help` advertise `start`, `test`, `test-monitoring`, `run-examples`, `update-slurm`, `test-all` — none of these targets exist in the file, and `slurm/README.md` tells users to run some of them.
- `slurm/Dockerfile:10`: pins `PYMCHELPER_VERSION=2.7.0` while `pyproject.toml` pins `pymchelper = "2.8.5"` for the backend that stamps `convertmc_version` into submitted scripts — worth aligning to avoid silent merge/parse mismatches.
- `slurm/docker-entrypoint.sh` (~line 91): slurmdbd-readiness loop passes the MySQL password on the command line (`mysql -h mysql -uslurm -ppassword`), visible to anything that can read the container's process list.
- `slurm/docker-entrypoint.sh` (~line 9): the `$SCRATCH` export/mkdir logic is duplicated verbatim into both `/etc/profile.d/scratch.sh` and `/etc/bashrc` instead of having the latter source the former.

---
*Automated critical review (Claude Code).*

PR #1662, opened 2026-08-05
Results pipeline rework (integration branch)

dazakdev:
Long-lived integration branch for the results pipeline rework. Work merges here; `master` only receives the finished result.


grzanka:
@dazakdev can you merge main/master branch so the tests could run ?

PR #1671, opened 2026-08-12
Numpy averaging

Mosder:
Changed unoptimal averaging which was using python list to instead use numpy. Also fixed a bug which would create a wrong average when a task failed


grzanka:
@Mosder take a look at the algorithm here: https://github.com/DataMedSci/pymchelper/blob/master/pymchelper/averaging.py

why don't you reuse it, it is numerically stable averaging (which includes also stddev/stderr calculations) with minimal memory footprint ?

